PR #1, opened 2021-11-11
Changes

ahasselbring:
Commits are sorted in order of presumed controversy, so you can cherry-pick some of them.


sseering:
Which editor do you use that parses `% !TeX spellcheck = en_US` comments?

ahasselbring:
TeXstudio

sseering (approved):
lgtm

Now that the terms "goal area" and "penalty area" are unified, we should also decide on one name for "penalty spot/mark/cross"

ahasselbring:
> Now that the terms "goal area" and "penalty area" are unified, we should also decide on one name for "penalty spot/mark/cross"

The IFAB/FIFA rules call it penalty mark. I guess the only reason that it is sometimes called penalty cross is to clarify its shape, but there is a sentence that explicitly says this plus the technical drawings.

Generally, conforming to IFAB/FIFA terminology (also sideline -> touchline) would also add another argument to the ongoing British/American English "edit war" :zany_face: (maybe I shouldn't have pushed the TeX spellchecking comments...)

PR #5, opened 2021-12-04
Changes, again

ahasselbring:
rebase of #1


sseering (approved):
_copypaste from the old PR_

lgtm

Now that the terms "goal area" and "penalty area" are unified, we should also decide on one name for "penalty spot/mark/cross"

PatrickGoettsch:
Me too

PR #9, opened 2021-12-15
Rules for the Visual Referee Technical Challenge

timewiley:
Rules for the visual referee technical challenge as pulled from the internal rule base.

Images for the hand-signals not yet completed. Expect these in a day or so.


content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+            \item Dynamic (motion) hand signals with a one or two hands.
+        \end{enumerate}
+        
+        The intent of this challenge is to choose a \emph{subset} of potential referee calls in SPL games and test ability of a robot to recognise different types of hand signals in preparation for adoption in RoboCup games, rather than compile a complete set of all referee signals.
 
     \subsubsection{Challenge setup}
 
-        One robot of the challenged team is placed in the center circle facing the head referee standing on the T-junction opposite to the GC. The referee wears red gloves. The robot watching the referee and listening for the whistle. The general procedure is as follows:
+        One robot of the challenged team is placed in the center circle facing the head referee, standing upright (stiffened) with both hands by its side.
+        %Two robots (from one team) are placed on the field, with one robot on the centre circle kick-off position  for a receiving team and one on the team's penalty spot. Both robots should be facing the kick-off spot.
```
PatrickGoettsch, line 88: 1 robot is already complicated enough for this year. Further extensions are possible

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+        %(Note that it is possible to run the challenge up to two teams at the same time. The referee should take care to balance the direction of hand signals in this case).
+        \todo{TW: Multiple robots could be used for this challenge, and the angle of the robot to the referee could make the challenge more difficult.}
+        The head referee must stand on the T-junction of the center field line, opposite to the GC. The referee must wear a ``black-and-white'' stripped referee jersey (if available) and must  wear red gloves. The purpose of this clothing is to clearly distinguish the referee and their hands from background people.
```
PatrickGoettsch, line 91: Instead of color can't you use the phrase left and right team out of the perspective of the head ref?

content/technical_challenges.tex
timewiley: Yes, I agree, this was an initial thought, but got called back. This comment can be removed.

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+        The general procedure is as follows:
 
-        \begin{itemize}
-            \item The referee blows the whistle.
-            \item The referee shows his decision for \qty{15}{\second}.
-            \item During that time or in the additional \qty{10}{\second} the robot has time to phrase the referee's decision, e.g., ```Kick-off left team'''. \todo{Or we use a TCP message}
-        \end{itemize}
+        \begin{enumerate}
+            \item The referee blows the whistle once. The referee may use one hand to hold (and blow) the whistle. The referee may instead leave the whistle held in their mouth without their hands.
+            \item The referee selects and indicates a hand-signal. The referee must indicate this signal for \qty{10}{\second}. \todo{reduced this time as 15s is unnecessarily long.}
```
PatrickGoettsch, line 99: Whistling and signaling should be done afterwards. So give a whistle. Wait a second then show the command.

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+            \item During that time or within an additional \qty{10}{\second}, the robot must indicate the hand-signal that it has identified by:
+            \begin{enumerate}
+                \item Using its hand(s) to \emph{mirror} the referee's signal. That is, if the referee given a signals for a decision for the ``red team'' using their right hand, the robot should \emph{mirror} this signal using the robot's \emph{left hand}, as the robot is facing the referee.
+                \item Providing an audio phrase the referee's decision, e.g., ```Goal Red Team'''.  The exact wording is up to the team, but should clearly identify the referee's signal.
+                \item (Optional), At the discretion of the GC/TCM developers, TCP messages may also be sent and interpreted to display the outcome of the robot's decision, however, this is not a requirement for the execution of the challenge. \todo{TW: Allowing this to be optional for now}
+            \end{enumerate}
```
PatrickGoettsch, line 105: Sounds reasonable.

content/technical_challenges.tex
timewiley: I have though about this. In our normal games we refer to colours of the teams, not directions of play. SO my goal here was to keep some parity for this challenge with games rules.

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+            \item The length of time taken for the robot to indicate it's interpretation from the referee blowing the whistle is measured (rounded-up to the nearest second).
+            \item If the robot cannot identify the signal, it should remain motionless and provide no audio output. A robot may continue to pose in the same position as the last signal it identified.
```
PatrickGoettsch, line 107: ... as the last signal IS identified

content/technical_challenges.tex
timewiley: Isn't that what's already written? Or do you want to have a explicit wait-1-second?

timewiley:
I've updated now with 'stick-figure' images. These aren't copied from anywhere so we won't have any copy write issues.

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+            \item While not providing a hand-signal or using the whistle, the head referee must keep both hands flat and motionless by their side.
+        \end{enumerate}
 
-        This procedure will be continued for in total four times. Each time, the referee chooses a new decision.
+        This procedure will be repeated \textit{five} times. Each time, the referee chooses a new decision. The referee should choose 2 one-hand static signals, 2 two-hand static signals, and 1 dynamic signal.
```
PatrickGoettsch, line 111: We could allow that the same signal can be chosen twice, so the set of options does not decrease. And your mentioned selection should be randomized.
It would be worth mentioning that the head ref and GC agree before the execution starts on order to be called. So, there are two people to listen to the robot and to support each other

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+            \item Static hand signals with two hands.
+            \item Dynamic (motion) hand signals with a one or two hands.
+        \end{enumerate}
+        
+        The intent of this challenge is to choose a \emph{subset} of potential referee calls in SPL games and test ability of a robot to recognise different types of hand signals in preparation for adoption in RoboCup games, rather than compile a complete set of all referee signals.
 
     \subsubsection{Challenge setup}
 
-        One robot of the challenged team is placed in the center circle facing the head referee standing on the T-junction opposite to the GC. The referee wears red gloves. The robot watching the referee and listening for the whistle. The general procedure is as follows:
+        One robot of the challenged team is placed in the center circle facing the head referee, standing upright (stiffened) with both hands by its side.
```
PatrickGoettsch, line 87: We should mention how to get the stiff robot from initial into challenge mode. With or without GC.

content/technical_challenges.tex
PatrickGoettsch: Okay

content/technical_challenges.tex
PatrickGoettsch: I think we should add the clearly to indicate that whistelling and signalling should be done after each other. From my experience a lot of referees do both at the same time.

PatrickGoettsch:
> 

They look nice

PatrickGoettsch:
Will you add the raw file to the repo as well to be able to create images in the future?

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+    \subsubsection{Available Hand-Signals}
 
-        For each decision (not all, because some decisions have to be shown on place, e.g., throw in) will be described and pictured.
+        Each hand-signal for the challenge is described and pictured. Note that for the purpose of clarity, these do not necessarily correspond to human soccer hand-signals.
 
         \begin{itemize}
-            \item \textbf{Kick-Off}
+            \item \textbf{Kick-in \textlangle{}colour\textrangle{} Team.}
+            One-handed signal. One arm, extended horizontally in the direction of the half of the field corresponding to the team that receives the Kick-in Free Kick. That is, right arm extended for the ``Blue team'', and left arm extended for the ``Red team''.
+            \todo{I'm not sure if this is consistent with human soccer, this could be inverted}.
```
PatrickGoettsch, line 120: The problem is that in normal soccer the referee is moving on the field. But we have here the assumption that he does not move. So, we must adapt the rules and signals for now. It could be a good extension for next year. To have a moving referee.

content/technical_challenges.tex
```diff
@@ -71,36 +71,136 @@ \subsection{Visual Referee Challenge}
...
+                  \includegraphics[height=130px]{figs/referee-signals/goal-kick.png}
+                \end{subfigure}
+                \begin{subfigure}{.33\textwidth}
+                  \includegraphics[height=130px]{figs/referee-signals/goal-kick-flipped.png}
+                \end{subfigure}
+            \end{figure}
+
+
+            \item \textbf{Corner Kick \textlangle{}colour\textrangle{} Team.}
+            One-handed signal. One arm, extended 45-degree \emph{down} in the direction of the end of the field where the corner kick will occur. That is, right arm extended for the ``Blue team'', and left arm extended for the ``Red team''.
```
PatrickGoettsch, line 149: Aren't here the teams reversed. Because corner kick happens for the red on the right half and for blue on the left half

content/technical_challenges.tex
timewiley: Ok, though in practice I don't think the explicit 1 second will make much difference, so long as the hand-signal _follows_ the whistle.

content/technical_challenges.tex
timewiley: yes

content/technical_challenges.tex
timewiley: yes to all.

timewiley: Thought to my understanding GC isn't required for this challenge (if the TCM packages are not being monitored).

content/technical_challenges.tex
timewiley: To my understanding GC isn't required for this challenge (if the TCM packages are not being monitored). So I would think in the event of no GC, the robot is simply enabled in the challenge mode by the team when placed. Nothing complicated or explicit needs to be done. If GC is used, then really it's just TCM, and the team can check the robot is connected.

timewiley:
> Will you add the raw file to the repo as well to be able to create images in the future?

They are in a Mac program, so I'm not sure they are useful to others. I can check photoshop compatibility at a later time.

content/technical_challenges.tex
timewiley: I'm not sure I understand this. To my knowledge the direction the ref points is always consistent to the signal. For example the side-ref uses the flag in a consistent direction for a thrown-in (and the field ref confirms the signal?).

content/technical_challenges.tex
timewiley: Ah yes. These should be interchanged.

content/technical_challenges.tex
PatrickGoettsch: I agree.

timewiley:
I've been travelling for end-of year. I've pushed changes now.

PatrickGoettsch:
Comment by Thomas Röfer:
However, I also have a question about that challenge. Does "one-handed signal" mean that it is undefined what the second hand does? "Kick-in" is a "one-handed signal", but that one hand is in the same position as during "goal" and "pushing". So the other hand decides which of the three signals is meant. Can we expect that, at least shortly after the whistle, the referee will have lowered the other arm as shown in the images?

timewiley:
> Comment by Thomas Röfer:
> However, I also have a question about that challenge. Does "one-handed signal" mean that it is undefined what the second hand does? "Kick-in" is a "one-handed signal", but that one hand is in the same position as during "goal" and "pushing". So the other hand decides which of the three signals is meant. Can we expect that, at least shortly after the whistle, the referee will have lowered the other arm as shown in the images?

I have pushed a new commit to clarify what to do with the non-signal hand in each one-handed signal. The result is to keep the hand by the ref's side.

PatrickGoettsch:
LGTM

PR #10, opened 2021-12-15
Champions Cup & Challenge Shield Split

timewiley:
Draft of split between the Champions Cup & Challenge Shield.
Many notes in the TODO.

I expect this to need a bit of a discussion


content/app_changes.tex
```diff
@@ -126,6 +127,20 @@ \subsection*{Official Rules - Disqualification during competition}
   \item Added possibility to disqualify a team during competition for serious ethical breaches, or violation of the terms of their qualification~(\cf \cref{sec:disqualification_during_comp})
 \end{itemize}
 
+\subsection*{Official Rules - Championship Cup and Challenge Shield}
```
PatrickGoettsch, line 130: We agreed on adding just the 5 to 10 major changes to the change list. Because the rest can be found in the repo

content/app_competition_rules.tex
```diff
@@ -47,18 +47,34 @@ \subsection{Qualification Procedure and Code Usage}
 
 \subsection{Game Structure}
 
-The clock stops during stoppages of play (such as ready and set state after goals) from the quarter-finals onward.  In round robin pool play, a game can finish in a draw as no penalty shoot-out will follow. In the promotion round, intermediate round, quarter finals, semi finals, 3rd place or final, a game that ends in a draw will be followed by a penalty shoot-out (see \cref{sec:penalty_shoot-out}).
+The clock stops during stoppages of play (such as ready and set state after goals) from the knock-out competition onwards.  In round-robin pool play, a game may finish in a draw and there will be no penalty shoot-out. Where a game cannot finish in a draw, including the knockout-competition (quarter finals, semi finals, 3rd place or final) and the promotion games, a game that ends in a draw will be followed by a penalty shoot-out (see Section~\ref{sec:penalty_shoot-out}).
```
PatrickGoettsch, line 50:
```suggestion
The clock stops during stoppages of play (such as ready and set state after goals) from the knock-out competition onwards.  In round-robin pool play, a game may finish in a draw and there will be no penalty shoot-out, like in all other games (see Section~\ref{sec:penalty_shoot-out}).
```
Just to shorten the text

content/app_competition_rules.tex
```diff
@@ -71,17 +87,132 @@ \subsection{Winner and Rankings}
...
+    \item All qualified teams that won the promotion game  of the previous year SPL RoboCup Champions Cup.
+    \item The highest Glicko ranked qualified team, until all remaining places are filled.
+\end{enumerate}
+
+All remaining qualified teams are then selected for the Challenge Shield.
+
+
+\subsection{Challenge Shield Rule Changes}
+\label{sec:cs-rule-changes}
+
```
PatrickGoettsch, line 111: My suggestion at this point would be to place the changes next to to corresponding rule in a box. This way, everybody sees that this rule differs whenbetween the two Leagues

content/app_competition_rules.tex
```diff
@@ -47,18 +47,34 @@ \subsection{Qualification Procedure and Code Usage}
...
+The clock stops during stoppages of play (such as ready and set state after goals) from the knock-out competition onwards.  In round-robin pool play, a game may finish in a draw and there will be no penalty shoot-out. Where a game cannot finish in a draw, including the knockout-competition (quarter finals, semi finals, 3rd place or final) and the promotion games, a game that ends in a draw will be followed by a penalty shoot-out (see Section~\ref{sec:penalty_shoot-out}).
 
-\subsection{Winner and Rankings}
-\label{sec:rankings}
+\subsection{Champions Cup and Challenge Shield}
+\label{sec:twoCompetitions}
+
+To provide better matched games for teams of all abilities the RoboCup Standard Platform League shall be divided into two separate competitions: the Champions Cup for the strongest teams and the Challenge Shield for all other teams. Champions Cup and Challenge Shield teams compete separately in each tournament. Teams are promoted to (or relegated from) the Champions Cup during the tournament.
```
PatrickGoettsch, line 55:
```suggestion
To provide better matched games for teams of all abilities the RoboCup Standard Platform League shall be divided into two internal leagues: League 1 for stronger teams and League 2 for all other teams. In both leagues teams compete separately in each league during the tournament. Teams from both leagues are promoted to or relegated from League 1 after all games in each league have been played and the promotion games take place. 
```
Regarding naming: I thought leagues would also be a good phrase. Like we have it in Germany or England. And I just see an issue with the promotion games. Because normally the last games played at RoboCup are the final games. This would mean that final games of League 2 would have to be executed much earlier to allow the promotion games to be executed before Final games of League 1 happen. Or they will play inbetween, somehow.

content/app_changes.tex
timewiley: OK...

content/app_competition_rules.tex
timewiley: If the TC is happy with the changes then, this can be considered. However, for the revival, it is easier to simply see everything all at once rather than trawl the rest of the rules.

content/app_competition_rules.tex
timewiley: I'm not going to change the names yet, without general agreement, so I've left this commit suggestion for now.

RE promotion game times: Once the bottom CC teams, and finalist CS teams are known the promotion games can really be scheduled whenever. The proposed changes have 5-6 more games than the current structure, but don't required the 3 entire rounds of play before the knockout. I would think it feasible to hold promotion games from the 2nd last day onwards (and on the final day in-between the generally scheduled final games). This also gives the bottom placed CC teams something to prepare for on the finals day(s). However, the final scheduling is up to the OC as games can be fit in.

content/app_competition_rules.tex
timewiley: I have now pushed a set of references (using todo notes for now) in the main body to where rules have been changed.

content/app_competition_rules.tex
```diff
@@ -71,17 +85,123 @@ \subsection{Winner and Rankings}
...
+
+
+\subsection{Challenge Shield Rule Changes}
+\label{sec:cs-rule-changes}
+
+To encourage new teams to join the SPL and reduce the bar to entry, the Challenge Shield will play under a modified rule-set to the competition rules (Sections \ref{sec:game_process} and \ref{sec:forbidden_act}). The intent of these rule modifications is to phrase a distinguishing factor between the Champions Cup and Challenge Shield as a higher level of refinement of a team's autonomous robot software through:
+\begin{itemize}
+    \item Reducing the harshness of penalties to allow robots to be on the field for a greater portion of the game time.
+    \item Give robots more time to react to different game circumstances.
+    \item Keeping consistency with the Champions Cup rules for teams that are promoted.
```
ahasselbring, line 114: Do I understand it correctly that promotion games are played according to champions cup rules?

content/app_competition_rules.tex
```diff
@@ -47,18 +47,32 @@ \subsection{Qualification Procedure and Code Usage}
 
 \subsection{Game Structure}
 
-The clock stops during stoppages of play (such as ready and set state after goals) from the quarter-finals onward.  In round robin pool play, a game can finish in a draw as no penalty shoot-out will follow. In the promotion round, intermediate round, quarter finals, semi finals, 3rd place or final, a game that ends in a draw will be followed by a penalty shoot-out (see \cref{sec:penalty_shoot-out}).
+The clock stops during stoppages of play (such as ready and set state after goals) from the knock-out competition onwards.  In round-robin pool play, a game may finish in a draw and there will be no penalty shoot-out, like in all other games (see Section~\ref{sec:penalty_shoot-out}).
```
ahasselbring, line 50: To which category do promotion games belong?

content/app_competition_rules.tex
```diff
@@ -71,17 +85,123 @@ \subsection{Winner and Rankings}
...
+The game changes to the \textit{Ready} state.
+This denotes that the robot's are given time to setup and prepare for the free kick.
+The game clock is \textit{paused} during finals games only.
+
+The head referees \emph{immediately} places the ball on the spot for the Free Kick.
+Players have \qty{\SCFreeKickSetupTime}{\second} to get into position for the Free Kick.
+At the end of \qty{\SCFreeKickSetupTime}{\second}, the game changes to the \textit{Set} state.
+Similar to a kick-off, during the Set state the robots are waiting for the Free Kick to commence.
+Standard penalties apply.
+Additionally, no player may touch the ball and defending players may not be within \FreeKickRadius of the ball.
```
ahasselbring, line 161: No exception for the goalkeeper within its own goal/penalty area?

timewiley: A good point, I forgot to add that - there should be an exception for the goal keeper. I will get on this tomorrow.

content/app_competition_rules.tex
timewiley: This I haven't fully decided yet, and are open to discussion. I think the promotion games give encouragement to both the CS and CC teams involved. I would probably propose holding it under the CS rule-set as it is a relaxed and more flexible rule set. I don't think it would put the CC teams at a disadvantage - if anything their "harder" ruleset would give them an advantage towards more reliable and quicker play. Where as the other way around significantly disadvantages the CS teams.

content/app_competition_rules.tex
timewiley: I think this goes with the discussion above → see there.

PatrickGoettsch:
We leave this PR for a moment open, due to the idea testing Swiss System at RC2022

Issue #3, opened 2021-11-24
Kick-Off


PatrickGoettsch:
Hi TC,

I happened to be looking at some details of kick-off and I noticed the following:

unless it has changed since the draft rules PDF was produced, kick-off is spelled a few different ways (kickoff, kick off, kick-off). The FIFA rules spell it kick-off so we should probably stick with that for consistency.
    the SPL rules for kick-off differ from the standard FIFA rules (which is OK) but it might be interesting to consider alignment of some aspects, specifically:
        Although it seems difficult to find a totally clear statement of this, my understanding of the FIFA rules is that all players of the attacking team must be inside their own half until the ball is in-play, i.e. supporting players may not cross into the opposition half until after the ball has been touched by the kicker and is in play. In SPL, I believe supporting players often cross the half way line before the kicker has touched the ball. Adding the restriction to stay in their own half prior to the ball being in play would be a little more realistic and not all that hard to implement I think.
        Until 2016 a FIFA kick-off had to be played forwards, but it can now be played in any direction.
        It is possible for the attacking team to score a goal directly from a FIFA kick-off. (It is not possible for them to score an own-goal directly from a FIFA kick-off - a corner is awarded in that case.) Adding this to SPL would be a more radical change, so this can be left for now if it seems too much. Allowing direct goals from kick-off might encourage slightly more interesting positioning from the defensive team however.
        In FIFA, the kicker cannot touch the ball a second time until another player has touched it - an indirect free kick is awarded if they do. Therefore dribbling directly from kick-off is not permitted. In SPL it may be interesting to enforce this (unless there is only one attacking outfield player on the pitch).

Best regards,
Rudi

sseering:
>     Although it seems difficult to find a totally clear statement of this, my understanding of the FIFA rules is that all players of the attacking team must be inside their own half until the ball is in-play, i.e. supporting players may not cross into the opposition half until after the ball has been touched by the kicker and is in play. In SPL, I believe supporting players often cross the half way line before the kicker has touched the ball. Adding the restriction to stay in their own half prior to the ball being in play would be a little more realistic and not all that hard to implement I think.

I agree with this idea. It's relatively easy to implement. Giving a more realistic image is important for the image of the league. I think we should adopt this change.

[...]

PR #11, opened 2021-12-15
Kick-off change towards Fifa rules

PatrickGoettsch:


PatrickGoettsch:
Solves #3  for now. Futher rules adjustments like Stefan suggested will be discussed next time within 
#11

PR #14, opened 2022-03-30
Clarification of Kick-Off Rule and penalties

PatrickGoettsch:


PatrickGoettsch:
@[HULKs] PatrickG the way the rules are written the attacking robot should be penalized if it touches the center line before hitting the ball. I think that's a bit too strict and doesn't allow any variation in doing the kick off except for cautiously playing forward. I don't think being that strict was intended. wink

ahasselbring:
Surely the intention is that the halfway line counts as part of the "opponent's half center circle", so it is allowed (because the robot doing the kick-off is allowed to go there)?

PatrickGoettsch:
I checked the FIFA rules and they explicitly exclude the kicking off robot: https://www.thefa.com/football-rules-governance/lawsandrules/laws/football-11-11/law-8---the-start-and-restart-of-play#:~:text=all%20players,%20except%20the%20player%20taking%20the%20kick%20off
I propose we change "Before the ball is in play, robots are not allowed to leave their own halves" to "Before the ball is in play, all robots except the single robot taking the kick-off are not allowed to leave their own halves" so that a kick-off to the side becomes possible again.

PR #17, opened 2022-03-30
Goalie only in its goal area allowed to be within freekickradius

PatrickGoettsch:


ahasselbring:
This is the same as removing the exception because a free kick cannot occur closer than 1m to the defending team's goal area.

PatrickGoettsch:
> 

Thanks for that hint, I addressed this point now by removing the exceptions.

PR #31, opened 2022-05-10
Adresses indirect kick-off goal by the defending team

PatrickGoettsch:


content/game_process.tex
```diff
@@ -105,28 +105,27 @@ \subsubsection{Invalid Goal}
...
 \label{sec:indirect_kick}
 
-From any restart in play (Kick-off, Kick-in, Free kick) except for Penalty kick, the attacking team may only score a goal via an indirect kick.
-%, providing they have at least 3 active robots on the field at the time the restart occurs (or is awarded in the case of a free kick).
-A robot may not score a goal from a direct kick, including via deflections.
-The ball must be deliberately played-at a second time (by either another robot, or the same robot) before a goal may be scored.
-A deliberate play at the ball includes successfully kicking the ball, dribbling the ball (and subsequently leaving possession of the ball), or the goalkeeper playing at the ball with its hands.
-If a robot plays the ball to itself, this means that the ball must leave a circular area of at least \qty{0.75}{\metre} the robot before the ball is played a second time and to be considered as an indirect
-kick.
+The attacking team may only score a goal via indirect kick from any restart into \textit{playing} from a Free kick except for a Penalty kick. This rule applies also for the defending team in case of unexecuted Kick-off and the ball is free.
```
arnemoos, line 113: Isn't it clearer to just state: 
**Any team** may only score a goal via indirect kick from any restart into \textit{playing} **(including Free kicks)** except for a Penalty kick. This rule applies also for the defending team in case of unexecuted Kick-off and the ball is free.

PatrickGoettsch: This would be another case and not allowing the defending team to score a direct goal if a goal kick has not been executed by the attacking team. Or also at the Kick in.

rvilling: Are the indirect requirements for attacking and defensive team at kickoff dealt with somewhere else? If not it still needs to be called out explicitly.
Are deflections considered indirect or not? If not, this should still be stated explicitly.

arnemoos: Ok, I thought the idea was to only allow indirect kicks. So no goal direct from the goal kick or some kick ins

PatrickGoettsch:
@rvilling 

> The text reads "A goal is invalid (that is, it can never be awarded) in the following circumstances: 1. When an Indirect Kick has not occurred (see Section 3.3.2)." This statement seems misleading as it implies that all goals must be indirect. Better to slightly reword as "1. When an Indirect Kick is required has not occurred (see Section 3.3.2)."

I addressed you point here

content/game_process.tex
PatrickGoettsch: There is a list which includes all the points directly below lines: 115 to 120

content/game_process.tex
```diff
@@ -100,33 +100,32 @@ \subsubsection{Invalid Goal}
 
 A goal is invalid (that is, it can never be awarded) in the following circumstances:
 \begin{enumerate}
-    \item When an Indirect Kick has not occurred (see~\cref{sec:indirect_kick}).
+    \item When an Indirect Kick is required has not occurred (see~\cref{sec:indirect_kick}).
```
ahasselbring, line 103: [...] is required *but* has not occured?
[...] *that* required has not occured?

content/game_process.tex
```diff
@@ -100,33 +100,32 @@ \subsubsection{Invalid Goal}
...
+\begin{description}
+  \item[Example 1:] Player 2 (of the red team) kicks the ball to Player~3, who then kicks the ball into the goal. This is a successful indirect kick, and the goal counts.
+  \item[Example 2:] Player 2 (of the red team) kicks the ball at the goal, and it is deflected of the side of the foot of a blue-team robot into the goal. This is \textit{not} an indirect kick, and the goal does not count.
+  \item[Example 3:] Player 2 (of the red team) kicks the ball ``upfield''. A blue-team robot kicks the ball a short distance, after which Player~2 kicks the ball again into the goal. This is a successful indirect kick, and the goal counts.
+  \item[Example 4:] Player 2 (of the red team), walks up to and dribbles the ball. To be an indirect kick, Player~2 must then stop, and visibly back-away from the ball, before approaching to dribble a \textit{second} time. The robot then scores. This is a successful indirect kick.
+  \item[Example 5:] Players of the red team misses to execute a Kick off. After the ball is free the defending team kicks the ball direct into the goal. This is \textit{not} an indirect kick, and the goal does not count.
```
ahasselbring, line 127: [...] of the read team *miss* to execute a *kick-off*. [...] kicks the ball *directly* into the goal. [...]

content/game_process.tex
rvilling: [...] is required but has not occurred

content/game_process.tex
PatrickGoettsch: Done

content/game_process.tex
PatrickGoettsch: Done

PR #36, opened 2022-05-10
Goal keeper handling if not directly on the line

PatrickGoettsch:


PatrickGoettsch:
this is the way we discussed in todays meeting

PR #41, opened 2022-05-10
Adds providing jersey in 7vs7

PatrickGoettsch:


content/technical_challenges.tex
```diff
@@ -27,6 +27,7 @@ \subsection{7 vs. 7}
             		\label{fig:initial_positions_7vs7}
             	\end{center}
             \end{figure}
+            Blue and red jerseys with numbers from 1 to 7 are provided by the OC/TC. If a team decides to play with own jerseys, they have to provide an additional one of the same shape.
```
arnemoos, line 30: an additional one **with the number 7** of the same shape.

PatrickGoettsch: Done

PR #47, opened 2022-05-11
Clarification of Kick off rule

PatrickGoettsch:


PatrickGoettsch:
Tries to close the potential gaps in the rules against abuse

content/game_process.tex
```diff
@@ -149,8 +149,8 @@ \subsection{Kick-off}
...
+In \textit{Ready} state, the robots must walk to a legal kick-off positions for their team.
+The kick-off team can be positioned anywhere within their own half including their half of the center circle. Only one robot of the kick-off team is allowed to stay within the whole center circle during \textit{Set} until the ball is in play. All robots except the closest to the center point of the center circle in the opponent's half of the center circle get penalized by ``Illegal Positioning''~(\cf \cref{sec:illegal_positioning}). This holds for robots who are positioned there wrongly in \textit{Set} or entering the opponent's half before the ball is in play. If the kick-off team has a robot positioned in the opponent's half of the center circle, only this robot, called kick-off player, is allowed to execute a kick-off. If another robot of the kick-off team is executing the kick-off than the kick-off player, this robot gets penalized with ``Illegal Positioning'' and a moved ball gets moved back. Even when the ball is free only all defending robots and the kick-off player are allowed to kick the ball. This restriction ends when the game state switches to \textit{Playing}.
```
arnemoos, line 153: All robots **in the opponent's half of the center circle,** except the closest to the center point of the center circle**,** get penalized by ``Illegal Positioning''~(\cf \cref{sec:illegal_positioning}).

arnemoos, line 153: If another robot of the kick-off team is executing the kick-off **than this** robot gets penalized with ``Illegal Positioning'' and a moved ball gets moved back.

Or should the Kickoff-Player be removed?

arnemoos (changes requested):

sseering:
I agree with the intent, but changed some of the language that I found ambigous or contradictory.

I changed the text and tried to improve it while keeping the intent. I included Arnes 2 comments into my changes.

I fully removed 

> Even when the ball is free only all defending robots and the kick-off player are allowed to kick the ball. This restriction ends when the game state switches to \textit{Playing}.

because I couldn't find a way to include it without contradictions. Overall, we could remove the whole "kick-off player" idea. This would simplify the rules and I'm not sure why we need to change it in the first place.

Fell free to make further changes or revert it if you don't like it.

content/game_process.tex
PatrickGoettsch: Done by Stefan

PatrickGoettsch:
LGTM

PR #51, opened 2022-05-11
Make in visual ref chal items to caption to align text with figures

PatrickGoettsch:


sseering (approved):
PDF looks good in my local build. All images are with their correct text.

PR #55, opened 2022-05-12
use more Si units quantities

PatrickGoettsch:


sseering (approved):
Local PDF build looks good

PR #56, opened 2022-05-12
Adds section about ref team communication

PatrickGoettsch:
This section introduces the idea to phrase the referee team communication and introduces a team captain. A team leader given word by RCF might not coincide with a responsible person during a match. 
A second point is that we get some randomness into the game as the side a team is starting and if it as kick-off is determined just 10 minutes before a game starts.
And it introduces polite communication during the games.


sseering:
Isn't it a bit late to introduce this method of switching starting halves? Do you see an urgent need here?

PatrickGoettsch:
Hi, no I do not see that particular point to be urgent. It could also be part of the next year. But the rest I think should be part. What do you think?

sseering:
> Hi, no I do not see that particular point to be urgent. It could also be part of the next year. 

I agree. Make is a proposal for next year.

> But the rest I think should be part. What do you think?

I agree with all other changes.

PatrickGoettsch:
I removed this part

content/judgment.tex
```diff
@@ -40,9 +40,19 @@ \subsection{Operator of the GameController}
...
 They are also responsible for keeping the time of each half (\ie, they stop the clock after a goal or game stuck, and continues it at the kick-off\footnote{The clock may not be stopped during the preliminaries.}).
-They should count aloud the remaining seconds in a half once the time remaining is 5 seconds or less.
+They should count aloud the remaining seconds in a half once the time remaining is \qty{5}{\second} or less.
 Finally, they should repeat the calls of the head referee to make sure it was heard correctly.
 
+\subsection{Referee — Team communication}
+\label{sec:referee_team_communiation}
+Both teams send a representative called team captain to the field \qty{10}{\minute} before their match starts. This time should be used for welcome each other, discuss match related topics.
+
+During a match only the team captains are communicating with the head referee. Only the team captains and two more people per teams are allowed to stay next to the game controller tables. The rest of the team keeps distance. This allows the referees easier communication with the team and the game controller operator gets less disturbed.
```
arnemoos, line 50: The rest of the team keeps distance **around the other sides of the field.**

content/judgment.tex
```diff
@@ -40,9 +40,19 @@ \subsection{Operator of the GameController}
 The operator will also inform the assistant referees when a timed penalty is over and a robot has to be placed back on the field.
 They should announce when the ball is in play on kick-off by stating ``Ball Free'', if the \KickOffBallFreeTime time period has elapsed in the playing state.
 They are also responsible for keeping the time of each half (\ie, they stop the clock after a goal or game stuck, and continues it at the kick-off\footnote{The clock may not be stopped during the preliminaries.}).
-They should count aloud the remaining seconds in a half once the time remaining is 5 seconds or less.
+They should count aloud the remaining seconds in a half once the time remaining is \qty{5}{\second} or less.
 Finally, they should repeat the calls of the head referee to make sure it was heard correctly.
 
+\subsection{Referee — Team communication}
+\label{sec:referee_team_communiation}
+Both teams send a representative called team captain to the field \qty{10}{\minute} before their match starts. This time should be used for welcome each other, discuss match related topics.
```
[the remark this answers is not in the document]
PatrickGoettsch: Done

content/judgment.tex
PatrickGoettsch: Thanks, got addressed

PR #82, opened 2023-01-10
Multiple robots in whole own half at kick-off

PatrickGoettsch:
Allow more than one play in whole own side

### DIFF
--- a/content/game_process.tex
+++ b/content/game_process.tex
@@ -30,11 +30,11 @@ If fewer robots do not respond to the GameController then they are, at the begin
 
   \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up. Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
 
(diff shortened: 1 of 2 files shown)


content/game_process.tex
```diff
@@ -163,7 +163,7 @@ \subsection{Kick-off}
 Robots are allowed to be positioned inside the area of their own goal.
 Both teams are also subject to restrictions on the penalty area~(\cf \cref{sec:illegal_positioning}).
 The limitations on kick-off positions end for both teams as soon as the ball is free.
-The green carpet border, except for the area of the goal, is not part of either teams own half.
+The green carpet border, except for the area of the goal, is not part of either teams' own half.
```
ahasselbring, line 166: I think the apostrophe must go before the genitive-s here.

arnemoos (approved):

content/game_process.tex
ahasselbring: Resolved in b93dd90

PR #84, opened 2023-01-10
removes indirect kick after kick off

PatrickGoettsch:
Not discussed in our last meeting but would be consequent to have this as well.

### DIFF
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -30,9 +30,8 @@ If fewer robots do not respond to the GameController then they are, at the begin
 
   \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up. Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
 
-  \item[Ready.] In this state, the robots walk to their legal positions for kick-off  (\cf \cref{sec:kick-off}) or a Penalty Kick (\cf \cref{sec:penalty_free_kick})). They remain in this state, until the head referee decides that there is no significant progress, up to a maximum of \qty{\KickOffAutoTime}{\second} for a kick-off and \qty{\PenaltyKickSetupTime}{\second} for a Penalty Kick.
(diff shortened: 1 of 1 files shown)


content/game_process.tex
```diff
@@ -170,10 +170,18 @@ \subsection{Kick-off}
...
-The defensive team must stay outside the center circle until the ball is in play. The ball is in play once it is touched by the attacking team or once \qty{\KickOffBallFreeTime}{\second} have elapsed in the playing state. Before the ball is in play, robots are not allowed to leave their own halves. Otherwise, a robot will be penalized with the ``Illegal Position'' penalty~(\cf \cref{sec:illegal_positioning}).
+The defensive team must stay outside the center circle until the ball is free. Before the ball is free, robots are not allowed to leave their own halves and additionally defensive players are not allowed to enter the center circle. Otherwise, a robot will be penalized with the ``Illegal Position'' penalty~(\cf \cref{sec:illegal_positioning}).
+
+A goal may be scored directly from a kick-off against the opposing team. If the ball goes directly into the goal of the player taking the kick-off, a corner kick is awarded to the opposing team.
+
+\subsubsection{Ball free}
+The ball is free and kick-off ends once 
+\begin{itemize}
+  \item it is touched by the attacking team and moved clearly or
```
ahasselbring: The "moved clearly" part is quite a difference to previous rules where people were often obsessed with the tiniest contacts between the foot and the ball (even more so in penalty kicks) that aren't even noticeable for the robots. I would actually be happy if this wasn't the case anymore (i.e. I support including the "moved clearly" part") because it is more consistent with real soccer (but then for free kicks and penalty kicks as well).

PatrickGoettsch: I agree except for penalty kick.

PatrickGoettsch: Is now in the PR

rules/game_process.tex
```diff
@@ -239,7 +239,7 @@ \subsection{Free Kick}
 
 A free kick is deemed completed and play returns to normal if:
 \begin{itemize}
-    \item The attacking team touches the ball, except for a robot getting up which is exempt from this rule.
+    \item The attacking team moves the ball clearly (except for a penalty kick: here is touching the ball sufficient to end the free kick), except for a robot getting up which is exempt from this rule.
```
ahasselbring, line 242: The penalty kick has its own section anyway.

PR #85, opened 2023-01-10
add coin toss to decide where to play and who has kick off

PatrickGoettsch:


content/game_process.tex
```diff
@@ -120,13 +118,23 @@ \subsubsection{Indirect Kick}
...
+  \item[Example 3:] Player 2 (of the red team) kicks the ball ``upfield''. A blue-team robot kicks the ball a short distance, after which Player~2 kicks the ball again into the blue team's goal. This is a successful indirect kick, and the goal counts.
+  \item[Example 4:] Player 2 (of the red team), walks up to and dribbles the ball. To be an indirect kick, Player~2 must then stop, and visibly back away from the ball, before approaching to dribble a \textit{second} time. The robot then scores. This is a successful indirect kick.
 \end{description}
 
+\subsection{Field-side selection and initial Kick-off}
+\label{sec:field_side_and_initial_kickoff}
+
+    \begin{itemize}
+        \item The head referee throws a coin with one side for the home team and the other side for the away team according to the schedule. The team that wins the coin toss can either choose which goal to play for in the first half or take the kick-off.
+        \item Depending on the decision above, the opposing team is given either the kick-off or may choose which goal to play for in the first half.
```
arnemoos, line 132: The winning seam can choose which side to play on or which team has kick off.
Subsequently, the second team can then decide if it wants to have kick off or can decide a side, but the team with the kick off selection has also the possibility to choose not to have kick off.

arnemoos (commented):
Currently your pull requests include all changes from the previous pull request was this so desired?

content/game_process.tex
```diff
@@ -6,15 +6,13 @@ \section{Game Process}
 \subsection{Structure of the Game}
 \label{sec:game_struct}
 
-A game consists of three parts, the first half, a half-time break, and the second half. Each half is \qty{10}{\minute} counted from the initial kick-off.
+A game consists of four parts, pregame meeting (see\,\ref*{sec:referee_team_communiation}), the first half, a half-time break, and the second half. Each half is \qty{10}{\minute} counted from the initial kick-off.
```
ahasselbring, line 9: pre-game, `\cref`.
Is there a reason to use `\,` here instead of `~`?

content/judgment.tex
```diff
@@ -55,14 +61,4 @@ \subsection{Referee -- Team communication}
...
-\subsection{A Remark on Artificial Landmarks}
-\label{sec:judgment:landmarks}
-
-The head referee may decide at any point before or during a game to relocate any objects around the field, or direct persons to another position around the field.
-
-The intent of using same-colored goals is to remove artificial landmarks.
-Robots should be able to localize with the SPL field and its ``normal'' surroundings.
-Introducing new team-specific artificial landmarks is against the spirit and intention of the league's progress.
-The application of this rule needs to be well considered and should be reserved for situations which seem constructed by one team or another, but will ultimately be the head referee's decision alone.
+The head referee and the assistant referees should wear socks \emph{of black or dark blue color} (blue jeans are acceptable) and avoid reserved colors for the ball, the goals, and player markings in their clothing. They may enter the field in particular situations, \eg, to remove a robot when applying a penalty. They should avoid interfering with the robots as much as possible.
```
ahasselbring, line 64: I support the idea that the trousers should not be as restricted as they used to be (-> "(blue jeans are acceptable)" should be removed, too). I would replace the "avoid reserved colors" part, too (the only left "reserved colors" are white and green).

content/judgment.tex
```diff
@@ -43,9 +49,9 @@ \subsection{GameController Operator}
 They should count aloud the remaining seconds in a half once the time remaining is \qty{5}{\second} or less.
 Finally, they should repeat the calls of the head referee to make sure it was heard correctly.
 
-\subsection{Referee -- Team communication}
+\subsection{Referee — Team communication}
```
ahasselbring, line 52: I think this should be `Referee--Team Communication`, but I can fix this in a later commit.

content/game_process.tex
PatrickGoettsch: > 

Yes, there where changes close by. So, I decided to do this. I do not know if it is the best way due to the dependencies.

PatrickGoettsch: Done

content/judgment.tex
PatrickGoettsch: Done

sseering:
I agree with the comment from Arne Moos from earlier: as written the rules don't allow the choice of not having kickoff in the first half.

Example: I want to have kickoff in the half where I start at the side of the field next to the glass wall. If my opponent wins the coin toss, they can choose to start in glass-wall-half for the first half. As written I would be forced to have kickoff in first half.  Thus I won't have kickoff with my team when starting in the glass-wall-half.

From out last meeting my understanding was that it should be allowed to choose to give the kickoff in the first half away.

PatrickGoettsch:
The point is, that the soccer rules are written as Arne has cited them. What I explained last Friday was wrong.

sseering (approved):

PR #91, opened 2023-01-12
In-game visual referee challenge 

PatrickGoettsch:


ahasselbring:
The least important aspect first: I think it should be spelled "in-game" with a hyphen.

challenges/visual_referee_challenge.tex
```diff
@@ -1,3 +1,163 @@
...
+    \end{subfigure}
+    \caption{\textbf{Player exchange.} Dynamic two-handed signal. Both arms slowly rotate symmetrically around a virtual horizontal axis parallel to the sideline. After three rotations the referee indicates with the kick-in signal the team, which is exchanging player.}
+\end{figure}
+
+\clearpage
+\newpage
+
+\subsection{Challenge evaluation}
+
+A team scores 1 point for every hand-signal that is correctly identified. A team scores an additional 1 point for correctly identifying the team corresponding to the signal (where appropriate: For a correct recognized \textit{Full-Time} signal without announcing a team, the team gets also 1 point for the team color). A team looses 1 point for not or incorrectly identifying a hand-signal (note a team may have a negative final score). The time for the robot to identify a hand-signal is timed (If a robot fails to identify a hand-signal the time for the hand-signal is \qty{15}{\second}. If a robot incorrectly identifies a hand-signal, the time is how long the robot took to provide the incorrect identification).
```
ahasselbring, line 150: This still has the same scoring flaw as last year (for the sided signals):

| | Side wrong | Side correct |
|---|---|---|
| Signal wrong | 0 | 0 |
| Signal correct | 1 | 2 |

i.e. you get partial points for [...], or e.g. if the actual signal is "kick-in red" and you identify "kick-in blue", you get 1 point even though they are *completely different* poses. On the other hand, if the actual signal is "goal kick red" and you identify "corner kick red", you get 0 points, although these signals are very similar.

Therefore, it would make more sense to be

| | Side wrong | Side correct |
|---|---|---|
| Signal wrong | 0 | 1 |
| Signal correct | 0 | 2 |

i.e. you should get partial points for identifying the side but not the signal. 

looses -> loses


Generally, the last two days have reinforced my opinion that things get pushed to master directly, since then I can also immediately fix them instead of trying to explain in a convoluted way what I mean.

sseering:
Fixes #80

challenges/visual_referee_challenge.tex
```diff
@@ -1,3 +1,163 @@
...
+
+During \textit{Set} and anytime during \textit{Playing} when the head referee whistles (Kick-off, goal, half ends), the challenged team (can be both teams) have to look to the T-junction opposite to the GameController operator. At the T-junction of the center field line a challenge assistant will show within \qty{5}{\second} after the whistle a random hand-signal and direction according to the list below for \qty{10}{\second}. The hand-signals and directions are randomly chosen but within the first four whistles there have to be hand-signals from all three classes. The same hand-signal may be chosen twice (with different directions). The challenge assistant is wearing red gloves to distinguish him from the other referees. In case the head referee has to stand on the T-junction, he will stand close to the challenge assistant and do his referee job from that close by position.
+
+The challenge assistant must wear referee cloth and red gloves. The purpose of this clothing is to clearly distinguish the challenge assistant and its hands from the other referees and from the background people.
```
sseering, line 25: Typo: "clothes", not "cloth"

sseering (changes requested):

challenges/visual_referee_challenge.tex
```diff
@@ -1,3 +1,163 @@
...
+
+The challenge assistant must wear referee cloth and red gloves. The purpose of this clothing is to clearly distinguish the challenge assistant and its hands from the other referees and from the background people. 
+
+The description of this challenge and hand-signals are described based on the viewpoint of the challenge assistant. In these descriptions from the perspective of the head referee the ``red team'' is defined as playing from left-to-right, and the ``blue team'' as playing from right-to-left. The use of colors for identifying teams is used to give equivalence to the head referee calls during the match. Blue and red have to be replaced by the actual colors of that particular match.
+
+The challenged team represented by the robot with the lowest number reports its evaluation of the particular hand-signal and direction using very loud sound output, repeating the result four times and sends a message with the decision over Wifi. \textbf{Details about this will come.} Only Messages with audio output and message sent within \qty{15}{second} after the signalling started, will be accepted.
```
PatrickGoettsch, line 29: Only Wifi, put into GameController return message.

PatrickGoettsch: Think about message budget for other challenge

PatrickGoettsch: Just take the first incoming message by a team, otherwise it is the one with the lowest number

PatrickGoettsch: a second whistle within 30s it has to be ignored

PatrickGoettsch: robot should send whistle, decision, team and timestamp

challenges/visual_referee_challenge.tex
PatrickGoettsch: Done

challenges/visual_referee_challenge.tex
PatrickGoettsch: Done

PR #97, opened 2023-01-18
We only play with GC

sseering:

In earlier sections of the rules its already stated that we only play with GC. We don't need special handling for chestbutton control anymore. So we can remove this.

Please discuss.


ahasselbring:
Yes.

PR #98, opened 2023-01-18
Reintroduce the primary states before listing them.

sseering:

The section begins with 

> Robots can be in eight different primary states.

Then 3 paragraphs of other text. 

Then the states are listed. 

I think this way it reads more nicely. Please discuss.


ahasselbring:
Yes.

PR #100, opened 2023-01-18
Don't block kickoff spot.

sseering:

So far we mention this edge case for the penalty kicks, but not for kickoff. We should mention it everywhere or nowhere. Please discuss.


ahasselbring:
Yes. I have some more local changes in this section but I can also apply them afterwards (e.g. "center point of the center circle" should be "center mark").

PR #104, opened 2023-01-27
Jersey Choice

ahasselbring:


ahasselbring:
Following up on the discussion about goalkeeper jerseys, I came up with this change:

> -\item Teams must have two sets of jerseys for \emph{goalkeepers} that are significantly different in terms of their primary color.  One of the \emph{goalkeeper} primary colors must be different from both primary colors used for \emph{field players}.
>+\item Teams must have two sets of jerseys for \emph{goalkeepers} that are significantly different in terms of their primary color.  At least one of the \emph{goalkeeper} primary colors must be different from both primary colors used for \emph{field players}.  If one of the \emph{goalkeeper} primary colors is similar to a \emph{field player} primary color, there must be a third \emph{goalkeeper} primary color (which may be identical to the other \emph{field player} primary color, though, which includes the possibility that they are in fact the same jerseys).

Basically this says that if a team wants to create only one additional goalkeeper jersey (set) (in addition to their previous colors), they must accept *both* of their previous colors as goalkeeper colors, too.

However, given the rules from this PR this may not even be necessary. I have come up with cases in which there is only a single possible jersey color assignment, but there seem to be no cases where there is none at all.

rules/robot_players.tex
```diff
@@ -53,7 +53,13 @@ \subsection{Team Markers}
...
+The jersey sets that are used in a game are decided by the head referee in accordance with the teams during the pre-game meeting (\cf~\cref{sec:referee_team_communication}), subject to the following conditions:
+\begin{itemize}
+  \item Both teams' \emph{field player} colors must be easily distinguishable from each other.
+  \item Both teams' \emph{goalkeeper} colors must be easily distinguishable from both teams' \emph{field player} colors.
+  \item The two \emph{goalkeeper} colors may only be similar when this is the case for all choices among the available jersey sets.
+\end{itemize}
+The head referee should respect the teams' wishes unless they violate the above conditions, in which case the head referee must bring about a solution.
```
PatrickGoettsch, line 62:
```suggestion
The head referee should respect the teams' wishes unless they violate the above conditions, in this case the head referee comes up with a solution.
```
Sounded a little bit weard for me. That is the only point here.

PatrickGoettsch:
I miss the commit that you mentioned before about the change in the color

rules/robot_players.tex
ahasselbring: I'm not 100% sure, but I think that "in which case" can join two clauses and "in this case" cannot. If you wanted "in this case", you'd have to split it into two sentences (or perhaps use a semicolon, but to me, a comma doesn't "feel" enough to separate the clauses). "in which case" is definitely a valid expression here.

For the second part, I agree that "bring about" is an uncommon phrase, as I didn't know it before I was looking for a translation of the German word "herbeiführen" (as in "der Schiedsrichter muss eine Lösung herbeiführen").

ahasselbring:
As I said, I don't consider this change necessary anymore, given that the referee can always find a solution even with the current rules (unless you find a counterexample).

PatrickGoettsch:
Okay. I see. Would be fine for me.

Issue #112, opened 2023-03-24
Data Minimization Challenge


ahasselbring:
- Since this challenge extends beyond the preliminary games: Should the time during penalty shoot-outs count?
- It's currently not specified whether the scored goal must be scored by a player of the team on the goal of the opponent (which I believe we agreed to - admittedly, it's hard to specify exactly what an own goal is vs. a deflection)

EmanueleMusumeci:
Regarding the first bullet point, I think that every data exchange during playing times should be taken into account, unless robots are penalized.

Regarding the second bullet, I think we should just refer to the score.

If none has anything to object, I will proceed with the required small integrations.

ahasselbring:
The rationale behind the first point is:
1. Team messages during penalty shoot-outs are not counted towards the message limit.
2. Therefore, it would be inconsistent to count messages in a penalty shoot-out towards the cumulated network traffic.
3. If it is not possible to increase the numerator of R^m during a penalty shoot-out, the denominator shouldn't increase either.
4. In penalty shoot-outs it's quite arbitrary when the state switches back from finished to set, where set counts as playing time (doesn't it?) and finished does not.
Therefore I think that including penalty shoot-outs in the challenge creates unnecessary bias that can be avoided by excluding them from the challenge.

EmanueleMusumeci:
Sorry, I thought my suggestion would make things easier for the GameController but it was actually complicating things :) I agree with your objection and I support the proposal of excluding penalty shoot-outs. Thank you for pointing this out!

PR #113, opened 2023-03-26
Update data_minimization_challenge.tex

EmanueleMusumeci:
Exclusion of penalty shoot-outs. I think the text as it is, is compatible enough with the second bullet point from Arne H. in issue #112 (https://github.com/RoboCup-SPL/Rules/issues/112)


ahasselbring (approved):

ahasselbring:
> I think the text as it is, is compatible enough with the second bullet point from Arne H. in issue https://github.com/RoboCup-SPL/Rules/issues/112 (https://github.com/RoboCup-SPL/Rules/issues/112)

It probably is compatible with that interpretation, but it turned out that readers who don't already know what is meant do not understand it that way.

PR #123, opened 2023-07-13
Revert sentence from 740673adcfe

ahasselbring:
The rewrite of this sentence in 2021 led teams to believe that transitions from finished to anything else can never happen, but they can due to the undo feature. This is exactly what the sentence was there for.

### DIFF
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -67,8 +67,9 @@ The primary states are:
   \centerline{\includegraphics[width=0.9\columnwidth]{figs/states_new.pdf}}
   \caption{Diagram of the robot states.
     \\\textbf{Chest button} transitions are shown as gray arrows.
-    However, any transition possible should be sent by the GameController.
     \\\textbf{GameController} transitions are shown as black arrows.
+    % The following sentence is necessary because the GameController can undo actions.
+    However, any transition possible can actually be sent by the GameController.
     \\Calibration transitions are shown as purple arrows which mean pressing the \textbf{front head button + the chest button}.
     \\From any state it can be transitioned to the \texttt{unstiff} state, shown as a blue arrow, via pressing all \textbf{three head buttons}.
     Pressing the \textbf{chest button} in the \texttt{unstiff} state allows a transition to the \texttt{initial} state, or a state as indicated by the GameController.}


rvilling:
"any transition possible" -- how are readers to know the possible transitions since they clearly include transitions not explicitly shown in the state machine (E.g. finished->playing or ready->playing in case of undo). Do you simply mean a transition from any to any state, or do you mean any transition shown in the state machine plus the undo (reverse traversal) of any of those transitions?

rvilling:
That sounds good.
________________________________
From: Patrick Gottsch ***@***.***>
Sent: Friday 14 July 2023 10:04
Cc: Rudi Villing ***@***.***>; Comment ***@***.***>
Subject: [EXTERNAL] Re: [RoboCup-SPL/Rules] Revert sentence from 740673adcfe (PR #123)

*Warning*

This email originated from outside of Maynooth University's Mail System. Do not reply, click links or open attachments unless you recognise the sender and know the content is safe.

Hi, I would phrase it differently: However, any transition possible should be sent by the GameController. Be aware that a transition can be reverted by the GameController operator by clicking the UNDO button.

I think that makes the situation clearer as it shows the usual transitions and makes the read aware that due to mistakes an tranition can be reverted and that they should handle this in their software.

ahasselbring:
I think that for a real solution all of this needs to be expressed in a different way that makes clear that the GameController generally does not report the true state of the game. The diagram is also incomplete because it doesn't show transitions from Ready/Set to Initial (timeouts), from Ready/Set to Finished (when time is over), from Finished to Initial (second half / penalty kick shoot-out), and from Finished to Set (penalty kick shoot-out). On the other hand, with those arrows the diagram would become more confusing than helpful.

>  by clicking the UNDO button
There is not really an undo button. I think the rule book should be agnostic of the GameController's UI.

ahasselbring:
> I don't think readers would find it helpful to read that the GC does not report the true state of the game. Is it not true that it reports the state signalled by the operator except for the specified delays (playing, goal)?

It is correct from the software perspective, but there is the question if the "true" state of the game is what the referee believes the state is, or what the GameController operator has entered. Going back to the example of the GameController operator pressing Finished in the middle of a half: Is the game really in the Finished state just because the GameController operator pressed the button? From a philosophical point of view it is not, because the head referee makes decisions, and ending the half was not a decision by the head referee. On the other hand, from a practical point of view the robots can't really know better (apart from looking at the remaining time).

Another example with actual consequences is when there is a free kick and shortly after, the referee calls an illegal defender, but the GameController operator has not entered the free kick yet. What is the true game state in this situation and is the referee's decision correct?

Anyway, I think your suggestion is good.

> As a final comment/question, is penalized really a primary state like the others in the figure?

No. I worked on that topic before RoboCup but didn't finish it. The current state is available here: https://github.com/RoboCup-SPL/Rules/tree/ahasselbring/future
I'm not sure if I'm going to continue to work on it, but I suggest the new TC to look into this.

Issue #106, opened 2023-01-31
Make Indirect Kick Rule realistic


ahasselbring:
At the moment, the indirect kick rule is a combination of two different concepts in real soccer:
- Indirect free kicks and throw-ins cannot directly score goals without another player touching it first. There is no differentiation between deflection and deliberate play (see [this video](https://www.youtube.com/watch?v=5725OukmQv4) for a throw-in goal that was only possible because the goalkeeper touched the ball).
- In all restarts of play, the player which brought the ball into play may not touch the ball again before it has touched another player. The rule book explicitly states for free kicks, throw-ins and corner kicks (here for the free kick):
  > If a player, while correctly taking a free kick, deliberately kicks the ball at an opponent in order to play the ball again but not in a careless or reckless manner or using excessive force, the referee allows play to continue.

  The same is not stated for kick-off, goal kick and penalty kick, but in those situations it doesn't really make sense for human players. Therefore, in my opinion, we should not have a differentiation between deflection and deliberate play either.

I propose to adopt this in the SPL in the following way:
- From any set play (kick-off, free kicks including kick-ins, goal kicks and corner kicks, penalty kick), the kicker may not touch the ball again before it has touched another player. *It does not matter if that touch was a deflection or an arbitrary play*. When touching the ball again before it has touched another player, an indirect free kick is awarded (I suppose we can save the direct free kick for a handball offense for the next decade). The kicker does not get a penalty (besides a free kick against its team). This restriction only holds for the attacking team and continues to hold even if the set play timed out (i.e. you can't circumvent it by waiting 10s after the kick-off - see below for another way to handle this).
- Kick-offs, penalty kicks, goal kicks and corner kicks are direct (i.e. a goal can be scored directly), while kick-ins and other free kicks (such as pushing free kicks or free kicks that occur due to the first rule) are indirect (i.e. they cannot score a goal, but *any* touch of another player is sufficient, i.e. it need not be deliberate play).

From there on, a few other changes could be made that go towards a more human-like game process:
- If a set play times out (e.g. 10 seconds after kick-off signal, 30 seconds after free kick, 30 seconds after penalty kick signal), instead of declaring the ball to be in play, a free kick could be awarded (this is justified because in real soccer, it is an offense to delay the restart of play). The fact that the ball can come into play without a player having touched it generally does not correspond to real soccer well.
[...]

PR #128, opened 2023-11-19
Adding more dynamic handling

ben47955:
Change to indirect kick rule

### DIFF
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -127,7 +127,7 @@ The head referee should also advise why the goal is invalid, such as by calling 
 \subsubsection{Indirect Kick - Champions Cup}
 \label{sec:indirect_kick_champions}
 
-Both teams may only score a goal by making a second contact with the ball within is own team from any affected start or restart into \texttt{playing}, never mind who initiate the kick: 
+Both teams may only score a goal through two distinct ball contacts involving different robots within its own team following any affected start or restart in the \texttt{playing} phase, regardless of the initiator of the set play/free kick:
 
 \begin{itemize}
   \item A robot may not score a goal from a direct kick, including via deflections.
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -124,17 +124,17 @@ A goal is invalid (that is, it can never be awarded) in the following circumstan
 In these cases a goal is not scored (that is, the goal is ruled invalid), and the game will proceed with a Goal Kick (\cf \cref{sec:free_kick}).
 The head referee should also advise why the goal is invalid, such as by calling ``Not indirect'', ``Played with hands'' or ``No own goal''.
 
-\subsubsection{Indirect Kick - Champion Cup}
-\label{sec:indirect_kick_champion}
+\subsubsection{Indirect Kick - Champions Cup}
+\label{sec:indirect_kick_champions}
 
-A team may only score a goal by making a second contact with the ball within is own team from any affected start or restart into \texttt{playing}. 
+A team may only score a goal by making a second contact with the ball within is own team from any affected start or restart into \texttt{playing}: 
 
 \begin{itemize}
   \item A robot may not score a goal from a direct kick, including via deflections.
-  \item Kick considered to be indirect are \texttt{Kick-off}, \texttt{Kick-in}, \texttt{Pushing free kick} and \texttt{Penalty kick}.
+  \item Set plays considered to be indirect are \texttt{Kick-off}, \texttt{Kick-in}, \texttt{Pushing free kick}, \texttt{Corner kick} and \texttt{Goal kick}.
   \item The ball must be deliberately played-at a second time by another robot of its own team before a goal may be scored.
-    A deliberate play at the ball includes successfully kicking the ball, dribbling the ball (and subsequently leaving possession of the ball), or the goalkeeper playing at the ball with its hands.
-  \item If an indirect free kick is kicked directly into the opponent's goal, the goal does not count and a goal kick is awarded.
+    A deliberate play at the ball includes successfully kicking the ball, dribbling the ball or the goalkeeper playing at the ball with its hands.
+  \item If an indirect free kick is kicked directly into the opponent's goal, the goal does not count and a and a \texttt{Goal kick} is awarded.
   \item Note that an own-goal may always be scored without requiring an indirect kick.
   \item Direct kick is allowed for the duration of the penalty kick, but this rule still applied after this kick. Contact in the penalty kick count among this rule.  
 \end{itemize}
@@ -143,18 +143,17 @@ A team may only score a goal by making a second contact with the ball within is 
   \item[Example 1:] Player 2 (of the red team) kicks the ball to Player~3 (of the red team), who then kicks the ball into the blue-team's goal.
   This is a successful indirect kick, and the goal counts.
   \item[Example 2:] Player 2 (of the red team) kicks the ball at the goal, and it is deflected off the side of the foot of a blue-team's robot into the goal.
-    This is \textit{not} an indirect kick, the goal does not count and a goal kick is awarded to the blue team.
+    This is \textit{not} an indirect kick, the goal does not count and a \texttt{Goal kick} is awarded to the blue team.
   \item[Example 3:] Player 2 (of the red team) kicks the ball ``upfield''.
     A blue-team robot kicks the ball a short distance, after which Player~2 kicks the ball again into the blue team's goal.
-    This is \textit{not} an indirect kick, the goal does not count and a goal kick is awarded to the blue team.
+    This is \textit{not} an indirect kick, the goal does not count and a \texttt{Goal kick} is awarded to the blue team.
   \item[Example 4:] Player 2 (of the red team), walks up to and dribbles the ball.
     Player~2 then stop, and visibly back away from the ball, before approaching to dribble a \textit{second} time.
-    The robot then scores. This is \textit{not} a successful indirect kick and a goal kick is awarded.
+    The robot then scores. This is \textit{not} a successful indirect kick and a \texttt{Goal kick} is awarded.
   \item[Example 5:] Player 2 (of the red team) kicks the ball ``upfield''.
   A blue-team robot kicks the ball again into the red team's goal.
   In this case, the blue teams does not need a indirect kick and the goal is counts.
-  \item[Example 6:] A penalty kick is awarded to the blue team. Player 2 (of the blue team) kick the ball. The red goalkeeper deflect the ball. Player 3 of blue team kick the ball in the red-team's goal.
-  This is a successful indirect kick, and the goal counts.
+  \item[Example 6:] A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick. A contact in the penalty kick count among this rule.
 \end{description}
 
 \subsubsection{Indirect Kick - Challenger Shield}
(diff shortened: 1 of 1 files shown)


rules/game_process.tex
```diff
@@ -127,28 +127,27 @@ \subsubsection{Invalid Goal}
 \subsubsection{Indirect Kick}
 \label{sec:indirect_kick}
 
-The attacking team may only score a goal via indirect kick from any restart into \texttt{playing} from a free kick.
+The attacking team may only score a goal via indirect kick from \texttt{Kick-off}, \texttt{Kick-in} and \texttt{Free Kick}.
```
arnemoos, line 130: Why specifically mention the Kick in? Wouldn't it not be clearer if we state: 
> The attacking team may only score a goal via indirect kick from any re-/start into \texttt{playing}.

rules/game_process.tex
```diff
@@ -127,28 +127,27 @@ \subsubsection{Invalid Goal}
...
 
 \begin{itemize}
   \item A robot may not score a goal from a direct kick, including via deflections.
-  \item The ball must be deliberately played-at a second time (by either another robot, or the same robot) before a goal may be scored.
+  \item The ball must be deliberately played-at a second time by another robot before a goal may be scored.
     A deliberate play at the ball includes successfully kicking the ball, dribbling the ball (and subsequently leaving possession of the ball), or the goalkeeper playing at the ball with its hands.
-  \item If a robot plays the ball to itself, this means that the ball must leave a circular area of at least \qty{0.75}{\metre} around the robot before the ball is played a second time and to be considered as an indirect kick.
+  \item if an indirect free kick is kicked directly into the opponent's goal, a goal kick is awarded.
```
arnemoos, line 136: _If_ an indirect free kick is kicked directly into the opponent's goal, _the goal does not count and_ a goal kick is awarded.

rules/game_process.tex
ben47955: That was my original plan, but then we need to exclude penalty kick from that. 
Fifa rules have a more specific list of of what kick is direct and indirect. I think for this year we should stay more generic to not add too much load on referees, but I thought this formulation was more in-lines with future evolution of this rule. That said, if the general opinion is to go with the simplified version, I'm fine with that.

rvilling:
Minor comment - the second competition is called "Challenge Shield" rather than "Challenger Shield"

rvilling:
In conventional usage (outside RoboCup) a free kick means a different thing to kick-in, corner kick, etc.

I'd prefer to see "The attacking team may only score a goal via indirect kick from any restart into \texttt{playing} from a free kick." rephrased to list the set plays from which an indirect kick is required (or to say all set plays except for some specified ones).

rvilling:
In the FIFA rules, the authors tend to list the full set of different set plays that applies in each case, e.g. "If the ball becomes defective at a kick-off, goal kick, corner kick, free kick, penalty kick or throw-in, the restart is retaken"

arnemoos:
@ben47955 maybe also have a look into issue #106

rules/game_process.tex
```diff
@@ -124,8 +124,41 @@ \subsubsection{Invalid Goal}
...
+    This is \textit{not} an indirect kick, the goal does not count and a goal kick is awarded to the blue team.
+  \item[Example 3:] Player 2 (of the red team) kicks the ball ``upfield''.
+    A blue-team robot kicks the ball a short distance, after which Player~2 kicks the ball again into the blue team's goal.
+    This is \textit{not} an indirect kick, the goal does not count and a goal kick is awarded to the blue team.
+  \item[Example 4:] Player 2 (of the red team), walks up to and dribbles the ball.
+    Player~2 then stop, and visibly back away from the ball, before approaching to dribble a \textit{second} time.
+    The robot then scores. This is \textit{not} a successful indirect kick and a goal kick is awarded.
+  \item[Example 5:] Player 2 (of the red team) kicks the ball ``upfield''.
+  A blue-team robot kicks the ball again into the red team's goal.
+  In this case, the blue teams does not need a indirect kick and the goal is counts.
```
arnemoos, line 155: This example seems wrong???

rules/game_process.tex
```diff
@@ -215,8 +248,6 @@ \subsection{Kick-off}
 The head referee must signal this from the T-junction of the halfway line.
 \qty{\PlayingDelayTime}{\second} after the head referee has signaled the kick-off, the robots' state is switched to \texttt{playing} by the GameController (regardless of when the ball comes into play).
 
-A goal may be scored directly from a kick-off.
```
arnemoos, line 218: A goal may be scored directly from a kick-off for the Challenge Shield. However, for the Champions Cup the indirect kick rule applies and the goal does \textit{not} count.

arnemoos (changes requested):

rules/game_process.tex
ben47955: My understanding of indirect kick is if the team don't do the kick, they don't need a indirect. While making indirect kick for both may make more dynamics, it's a step away from the fifa rule. Let me know if I'm still wrong.

rules/game_process.tex
```diff
@@ -124,8 +124,41 @@ \subsubsection{Invalid Goal}
...
+A team may only score a goal by making a second contact with the ball within is own team from any affected start or restart into \texttt{playing}. 
+
+\begin{itemize}
+  \item A robot may not score a goal from a direct kick, including via deflections.
+  \item Kick considered to be indirect are \texttt{Kick-off}, \texttt{Kick-in}, \texttt{Pushing free kick} and \texttt{Penalty kick}.
+  \item The ball must be deliberately played-at a second time by another robot of its own team before a goal may be scored.
+    A deliberate play at the ball includes successfully kicking the ball, dribbling the ball (and subsequently leaving possession of the ball), or the goalkeeper playing at the ball with its hands.
+  \item If an indirect free kick is kicked directly into the opponent's goal, the goal does not count and a goal kick is awarded.
+  \item Note that an own-goal may always be scored without requiring an indirect kick.
+  \item Direct kick is allowed for the duration of the penalty kick, but this rule still applied after this kick. Contact in the penalty kick count among this rule.
```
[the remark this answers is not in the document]
ben47955: I still need to think about that as it's kinda weird to talk indirect for penalty as it's not listed in the list.

ben47955:
I made most change, I juste need to solve the two points in pending

rules/game_process.tex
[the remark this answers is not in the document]
arnemoos: From the Penalty kick rules:
> The attacking robot (taking the penalty kick) may only score a goal if it touches the ball once.
> Once the attacking robot has touched the ball, it may not score a goal until another robot (from either team) touches the ball.

Maybe we can add the penalty kick in the list as: "or after an unsuccessful Penalty kick" or something in this way.

rules/game_process.tex
arnemoos: In my understanding, both teams need an indirect kick and this example is wrong. Also I agree that it might not fully align with the fifa rules I would say that this would again be a special case and makes it harder for the referee. So for this year I would say we always require from both teams to do an indirect kick.

rules/game_process.tex
```diff
@@ -124,8 +124,40 @@ \subsubsection{Invalid Goal}
...
+\label{sec:indirect_kick_champions}
+
+A team may only score a goal by making a second contact with the ball within is own team from any affected start or restart into \texttt{playing}: 
+
+\begin{itemize}
+  \item A robot may not score a goal from a direct kick, including via deflections.
+  \item Set plays considered to be indirect are \texttt{Kick-off}, \texttt{Kick-in}, \texttt{Pushing free kick}, \texttt{Corner kick} and \texttt{Goal kick}.
+  \item The ball must be deliberately played-at a second time by another robot of its own team before a goal may be scored.
+    A deliberate play at the ball includes successfully kicking the ball, dribbling the ball or the goalkeeper playing at the ball with its hands.
+  \item If an indirect free kick is kicked directly into the opponent's goal, the goal does not count and a and a \texttt{Goal kick} is awarded.
```
arnemoos, line 137: double "and a"

rules/game_process.tex
```diff
@@ -124,8 +124,40 @@ \subsubsection{Invalid Goal}
...
+  \item[Example 4:] Player 2 (of the red team), walks up to and dribbles the ball.
+    Player~2 then stop, and visibly back away from the ball, before approaching to dribble a \textit{second} time.
+    The robot then scores. This is \textit{not} a successful indirect kick and a \texttt{Goal kick} is awarded.
+  \item[Example 5:] Player 2 (of the red team) kicks the ball ``upfield''.
+  A blue-team robot kicks the ball again into the red team's goal.
+  In this case, the blue teams does not need a indirect kick and the goal is counts.
+  \item[Example 6:] A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick. A contact in the penalty kick count among this rule.
+\end{description}
+
+\subsubsection{Indirect Kick - Challenger Shield}
```
arnemoos, line 159: Challenge Shield

rules/game_process.tex
rvilling: Why do you think that both teams need (or specifically that the Blue team needs) an indirect kick for example 5. I wouldn't agree with both teams requiring an indirect kick. Before the kick was taken there was no requirement for the blue team to do an indirect kick and this should not change as a result of what the red player did.

Example 3 is also problematic. I would have thought that as soon as possession transferred to the team that was defending the kick, we're back to normal playing. I understand that you are probably trying to catch the case where the red player executing the indirect kick just kicks at the blue opponent and then immediately follows up the rebound to shoot and score. However, example 3 creates questions for referees. How long is "short kick" by the blue player and when can we consider that the indirect kick (for red) is over and we're back to normal play?

arnemoos: In my understanding after each start or restart into playing each of the teams has to do the indirect kick in order to score a goal which means for each team 2 different robots of the team have to touch the ball. So in this example it's not stated who has touched the ball. If for both teams already two different robots have touched the ball then this example holds. Otherwise not.

rvilling: I don't think that is obvious is in the rules and it is certainly not intuitive. Normally any rules about indirect free kicks or indirect kicks apply only to the team awarded the kick. If your interpretation is as you intended then it would need to be stated more clearly in the rules that this is what is meant.  E.g. line 130 should be extended to explicitly say that the requirement for indirect kicks applies to both the team who has kickoff and the team defending the kickoff.

rvilling: OK. This is up to the TC. Just make it really clear and explicit in the rules (not just by an example) - I think line 130 is the one to modify.

jpha226: I agree that a global rule for both teams is easier for the refs and I am in favor of that. 
However, it seems like the requirement is then saying that no one can score until *both* teams have had a pass. I don't think that is what we want. I think it should be that no one can score until at least one team has two different robots touch the ball.

ben47955:
I made the discussed change

rules/game_process.tex
```diff
@@ -124,8 +124,42 @@ \subsubsection{Invalid Goal}
 In these cases a goal is not scored (that is, the goal is ruled invalid), and the game will proceed with a Goal Kick (\cf \cref{sec:free_kick}).
 The head referee should also advise why the goal is invalid, such as by calling ``Not indirect'', ``Played with hands'' or ``No own goal''.
 
-\subsubsection{Indirect Kick}
-\label{sec:indirect_kick}
+\subsubsection{Indirect Kick - Champions Cup}
+\label{sec:indirect_kick_champions}
+
+Both teams may only score a goal by making a second contact with the ball within is own team from any affected start or restart into \texttt{playing}, never mind who initiate the kick:
```
arnemoos, line 130: Both team may only score a goal through two distinct ball contacts involving different robots within its own team following any affected start or restart in the \texttt{playing} phase, regardless of the initiator of the set play/free kick.

ben47955:
Updated

PR #129, opened 2023-11-21
Simplify standard penalty

tarandeep-panesar:


arnemoos:
I think what's still missing is where the robots should be placed behind the sideline (and on which side)?
Also we should add a precaution sentence that the GC should at least shortly check that the robot is really standing near the sideline :P

Issue #24, opened 2022-04-17
GC connection 


PatrickGoettsch:
After gore: new rule: a robot of a remote team does not listen to the GC then it is ejected from the game

ahasselbring:
What exactly is the difference to non-remote teams' robots? That they can't pick up the robot, change the software, and let it reenter?

rvilling:
Why should the rules differ for a remote team in this case? Is the challenge and responsibility not identical for local and remote teams?

As a minor detail, how will you decide if a robot does not listen to the GC as opposed to that it does listen but has some other bug in a particular state? If this is something that only applies in the ready state or on return from penalty (e.g. after request for pickup) then it makes a bit more sense to me.

PatrickGoettsch:
Does not apply to RC2022

PR #132, opened 2023-12-11
Resolve GC Connection Issues #24

arnemoos:
Will probably be handled after the draft release of the rules.


ben47955:
This seem to address a part of the initial issue. However, I'm not able to find a definition for a robot connected or not to GC ( Status green ? ) 
From there, what do we do if the robot isn't responding, but behalf correctly ?

arnemoos:
Ok: I [clarified what "not responding" means](https://github.com/RoboCup-SPL/Rules/pull/132/commits/af55f6d414c8680ecf1b2d82db235a07c2541a76)

ben47955:
I expressed some concern in the last meeting about this PR, but after some thought and it seems fine for me now.

PR #137, opened 2023-12-15
more detailed explanation of CC/CS split

sseering:
more detailed explanation of CC/CS split

### DIFF
--- a/rules/app_competition_rules.tex
+++ b/rules/app_competition_rules.tex
@@ -59,10 +59,10 @@ Matches in the champions cup are played with up to 7 players on each team, while
 
 In addition, some rules are simplified for Challenge Shield matches. The major differences are:
 \begin{itemize}
-  \item The requirement to recognize visual referee signals are relaxed for the Champions Cup (\cf~\cref{sec:robot_players}).
+  \item The requirement to recognize visual referee signals is relaxed for the Champions Shield (\cf~\cref{sec:robot_players}).
 
   % uncomment this after the relevant pull request is merged and update cref
-  % \item Goals may be scored via direct kick from kick-off. The indirect kick rules are relaxed for the Champions Cup (\cf~\cref{sec:TODO}).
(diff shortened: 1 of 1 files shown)


rules/app_competition_rules.tex
```diff
@@ -57,6 +57,14 @@ \subsection{Champions Cup and Challenge Shield}
 Teams are assigned to one of the competitions during the application phase, considering their preference, ability and organizational aspects.
 Matches in the champions cup are played with up to 7 players on each team, while the challenge shield is played with up to 5 players on each team (\cf~\cref{sec:robot_players}).
 
+In addition, some rules are simplified for Challenge Shield matches. The major differences are:
+\begin{itemize}
+  \item The requirement to recognize visual referee signals are relaxed for the Champions Cup (\cf~\cref{sec:robot_players}).
```
rvilling, line 62: "relaxed for the Champions Cup" seems to suggest that visual ref signals are not as strict in the CC. Did you mean to say that the rules for CS are  "relaxed relative to the Champions Cup" instead?

If so, you should adjust the commented out text in line 65 also

sseering: Thank you. I wrote the opposite of what I meant by accident. I changed the wording.

rules/app_competition_rules.tex
```diff
@@ -57,6 +57,14 @@ \subsection{Champions Cup and Challenge Shield}
 Teams are assigned to one of the competitions during the application phase, considering their preference, ability and organizational aspects.
 Matches in the champions cup are played with up to 7 players on each team, while the challenge shield is played with up to 5 players on each team (\cf~\cref{sec:robot_players}).
 
+In addition, some rules are simplified for Challenge Shield matches. The major differences are:
+\begin{itemize}
+  \item The requirement to recognize visual referee signals is relaxed for the Champions Shield (\cf~\cref{sec:robot_players}).
```
rvilling, line 62: Champions Shield -> Challenge Shield

rvilling: Looks fine now

Issue #121, opened 2023-06-14
Simplify Global Game Stuck


ahasselbring:
Many referees have problems to remember on which team to call Global Game Stuck, in part because there is negation involved (the team that is *not* closest to the ball, but the *other* team, i.e. the one with the closest player, gets the kick-off).

Although it shares the problem that the team getting the kick-off is not the one on which the Global Game Stuck is called (but this is also the case for goals), I propose to call the Global Game Stuck on the team in whose half the ball is (Global Game Stuck cannot occur as long as the ball has not been moved from the center mark, and the handling the case in which the ball is exactly on the halfway line is not fundamentally worse than deciding which player is closer to the ball in some situations).

PR #150, opened 2024-02-27
Global game stuck simplify

sseering:
See https://github.com/RoboCup-SPL/Rules/issues/121

### DIFF
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -420,7 +420,7 @@ The referee calls ``Global Game Stuck'' to announce the decision.
 
 Once the referee calls Global Game Stuck, players enter the \texttt{ready} state, and a new kick-off is awarded.
 For the first Global Game Stuck of a half, the kick-off is awarded to the team that did not have the initial kick-off of the half.
-For any further Global Game Stuck calls of a half, kick-off is awarded to the team that did not receive kick-off in to the previous Global Game Stuck call.
+For any further Global Game Stuck calls of a half, kick-off is awarded to the team that did not receive kick-off in the previous Global Game Stuck call.
 
 A global game stuck can only be called if at least one robot has touched the ball since the previous kick-off.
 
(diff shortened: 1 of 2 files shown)


rules/game_process.tex
```diff
@@ -415,10 +415,13 @@ \subsubsection{Local Game Stuck}
...
 \label{sec:game_stuck:global}
 
-If no robots have made progress towards the ball or began to explore the field in \qty{\GameStuckTime}{\second} Global Game Stuck should be called on the team whose robot is \textit{not} nearest the ball.
-The referee calls ``Global Game Stuck \textless color\textgreater''.
+If no robots have made progress towards the ball or began to explore the field in \qty{\GameStuckTime}{\second} Global Game Stuck may be called.
+The referee calls ``Global Game Stuck'' to announce the decision.
+
+Once the referee calls Global Game Stuck, players enter the \texttt{ready} state, and a new kick-off is awarded.
+For the first Global Game Stuck of a half, the kick-off is awarded to the team that did not have the initial kick-off of the half.
+For any further Global Game Stuck calls of a half, kick-off is awarded to the team that did not receive kick-off in to the previous Global Game Stuck call.
```
arnemoos, line 423: Can't this be shortened: 

The kick-off is awarded to the team that did not have the last (initial) kick-off. 

Also I would add that this is automatically determined by the GC.

arnemoos (changes requested):

rules/game_process.tex
sseering: As disucces on the meeting:

> Also I would add that this is automatically determined by the GC.

Will be added.

> The kick-off is awarded to the team that did not have the last (initial) kick-off.

We won't simplify to Global Game Stuck to keep the kickoff team independent form goals being scored.

arnemoos (approved):

ahasselbring:
Is this affected by timeouts?

arnemoos:
> Is this affected by timeouts?

No, only the initial kick-off and then global game stuck calls are changing it.

Issue #152, opened 2024-04-26
Add game state between "initial" and "ready"


roefer:
According to the new rules, the referee signals the beginning of the game with a gesture. While we managed to implement this feature and it always worked at the RoboCup German Open 2024, it makes the phase before the game more complicated. The problem is that if teams place their robots early on the field, they risk that the robots might detect a false gesture by accident, e.g. if someone is setting up a camera on a very long tripod right a the halfway line, lifting up both arms in the process, only half a meter behind the position where the referee would later show the gesture. Theoretically, teams could rely on the red gloves, but since it was already announced that they might not be used anymore in the future, it does not make sense to just integrate them for this year in the detection.

Our solution was to place the robots on the field, check whether they can see the ball, let them sit down afterwards, and letting them get up again when the referee arrives, so that they only then start to wait for the gesture. It would be lot easier, if the GameController signaled when the referee and the GameController operator are present and the game will start soon (basically a "set" state between "initial" and "ready").

With the current rules, it can happen (and it actually happened for at least one team) that the robots start moving by accident, but the referees are not yet ready to penalize them for "forbidden motion in initial". This should be avoided.

sseering:
I agree and others. Others during the German Open and me  thought of a similar changes. I talked to Arne Hasselring. Accoeding to him, the effort to implement it in the GameController would be not too big. 

In my Idea the addition game state would be before "initial" (so teams don't have to insert an new gamestate in the middle of the gamstate-transition-code, making the new gamestate less error prone to implement?). It could be called "pre-initial" or "setup-time" or similar.

sseering:
Dear Teams, a small heads up: due to feedback from the German Open we will be introducing a new GameState in the rules and GameController before the 2024 World Championship.

The new GameStatse will likely be called "Setup" and happen before the Initial in match. It gives teams time to set up on the side on the field. It will be the same as Initial (thus, easy to implement), but the robots will not be penalized for the "Motion in Initial" penalty.

We are changing the rules soon and will publish a new GameController (thanks to AH). We will notify as soon as the details are set in stone.

rvilling:
Hi Stefan,
are all teams subscribed to the github rules repo? I wonder if your message below should be sent directly to the participating teams (is there a mailing list?) or even just notified on discord.
Best regards,
Rudi
________________________________
From: sseering ***@***.***>
Sent: Tuesday 30 April 2024 11:29
To: RoboCup-SPL/Rules ***@***.***>
[...]

PR #153, opened 2024-05-02
Adding standby state before ready

ben47955:


ben47955:
Implementing #152

rules/game_process.tex
```diff
@@ -75,6 +78,8 @@ \subsection{Robot States}
   \label{fig:robot_states}
 \end{figure}
 
+The referee must tell the GameController operator to switch to \texttt{initial} state once everyone is in place or that teams are falling under forfeit rule (see \cref{sec:forfeit}).
```
sseering, line 81: "The referee will announce the transition from the \texttt{setup} to \texttt{initial} ...."

Make the sentence simpler and consistent with the following text 2 sentences later.

sseering, line 81: > once everyone is in place or that teams are falling under forfeit rule (see \cref{sec:forfeit}).

I don't like introducing "once everyone is in place" here.

The head ref starts the game according to the schedule. If a team is not ready in time its their own fault. What happens if you are late/no-show is already detailed elsewhere.

rules/game_process.tex
```diff
@@ -75,6 +78,8 @@ \subsection{Robot States}
...
+The referee should wait between 10 to 120 seconds before announcing the transition to \texttt{ready}.
```
sseering, line 82: How about increasing the mininmum to 30 secs?  

Other opinions?

rules/game_process.tex
```diff
@@ -89,7 +94,8 @@ \subsection{Robot States}
 The colors corresponding to the game states are:
 \begin{itemize}
   \item \texttt{Unstiff}: Blue-Blinking
-  \item \texttt{Initial}: Off
+  \item \texttt{Setup}: Off
+  \item \texttt{Initial}: Cyan or Off
```
sseering, line 98: Don't make the rules more complicated by introducing options here. Just leave it as off (the head ref can see the gamestate on the GSV/TCM screen anyways)? Other opinions?

rules/game_process.tex
sseering, line 81: Should we explain here that both halves start with the setup state? 

Previous sections already imply it, though the text is kinda vague. In the past the rules also didn't write more explicitly that halves start with initial. 

opinions?

sseering, line 81: We should use this change to formalize/define when human members of the playing teams need to leave the field/stop touching their robots. 

In the past it varied wildly from team to team how late before initial human members of the playing teams where touching the robots. It should be somewhat consistent across all teams for fairness.

We could introduce a requirement like "The head referee  announces setup time will be ending in 60 seconds. All human members of the playing teams must leave the field and stop interacting with participating robots manually and via network before initial."

rules/forbidden_actions.tex
```diff
@@ -112,7 +112,7 @@ \subsubsection{Motion in Initial}
 
 Robots that begin moving before the transition from \texttt{initial} to \texttt{ready} (\ie before the head referee does the visual signal).
 The head referee will call ``Forbidden Motion in Initial\textless robot\textgreater''.
-Note that responding to any false signal of any fashion will result in this penalty. This rule does apply when the head referee believes that both teams are ready.
+Note that responding to any false signal of any fashion will result in this penalty.
```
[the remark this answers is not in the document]
ben47955: Agree. I will update

rules/game_process.tex
```diff
@@ -33,12 +33,15 @@ \subsection{Robot States}
     So while in the \texttt{unstiff} state the robot is not allowed to move in any fashion! After booting, the robots are in their \texttt{unstiff} state.
     Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the \texttt{initial} state, or a state as indicated by GameController.
 
-  \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
-    Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
+    \item[Setup.] The robots are not allowed to be moving in any fashion besides initially standing up.
```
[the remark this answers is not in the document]
ben47955: I think the issue is the initial was not well defined and I did the same mistake. I will update.

rules/game_process.tex
ben47955: I didn't like either. I will try to formulate that differently.

rules/game_process.tex
ben47955: I didn't want to introduce a different rule for CS and CC. I think this should be handled at referee meeting. I'm sure no one really gonna wait for 120 seconds doing nothing, but at the same time we don't want every initial state to be 30 seconds. I think this give more room to have the duration between 20-45 seconds.

rules/game_process.tex
ben47955: I make it optional for now because we are late in the season, but at the same time I think it's great to the option to differentiate every state on the robot.  Suggestion for next years should be to make it mandatory.

rules/game_process.tex
sseering: If we handle this during the referee meeting, then we could not give any time in the rules. Keeping it simpler.

rvilling:
It would be important to signal to readers of the rules that there would be a delay to be finalised in the referee meeting so that no teams write their code on the expectation of no delay.
________________________________
From: sseering ***@***.***>
Sent: Friday 10 May 2024 09:29
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [RoboCup-SPL/Rules] Adding state before initial (PR #153)

*Warning*

This email originated from outside of Maynooth University's Mail System. Do not reply, click links or open attachments unless you recognise the sender and know the content is safe.

@sseering commented on this pull request.

________________________________

In rules/game_process.tex<https://github.com/RoboCup-SPL/Rules/pull/153#discussion_r1596446810>:

> @@ -75,6 +78,8 @@ \subsection{Robot States}
   \label{fig:robot_states}
 \end{figure}

+The referee must tell the GameController operator to switch to \texttt{initial} state once everyone is in place or that teams are falling under forfeit rule (see \cref{sec:forfeit}).
+The referee should wait between 10 to 120 seconds before announcing the transition to \texttt{ready}.


If we handle this during the referee meeting, then we could not give any time in the rules. Keeping it simpler.

ben47955:
Hello, I made change to the PR

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
   \item[Unstiff.] It helps to facilitate a consistent and safe handling of the robots for remote competition.
     During any state, if all head buttons are pressed at least one second, the robot should move to a safe seated/crouched position and unstiffen all joints.
     So while in the \texttt{unstiff} state the robot is not allowed to move in any fashion! After booting, the robots are in their \texttt{unstiff} state.
-    Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the \texttt{initial} state, or a state as indicated by GameController.
+    Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the state indicated by GameController.
```
rvilling, line 34: It looks like you're assuming there will always be a game controller already running. This isn't usually the case, e.g. if you arrive at a field before GC is running or if you don't want to run GC (E.g. to do calibration). The transition from a button press in unstiff seems undefined when there is no GC. Can you define what state the robot should land in if there is no game controller running (I assume "setup").

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
 
-  \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
+    \item[Setup.] The robots are free of any movement at teams convenience.
```
rvilling, line 36: This wording is a bit unclear. Do you mean that the robots can move in the setup state or that they cannot?

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
+    This state is not limited in time and teams have access to the field.
+    The GameController will activate this state before \texttt{initial} (i.e at the beginning of a half and during a timeout).
+
+    \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
```
rvilling, line 40: "to be moving" --> "to move"

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
     Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
+    The robots are preparing to receive a visual signal from the referee.
```
rvilling, line 42: Suggest rewording: In \texttt{iniitial}, the robots are awaiting a visual indication from the referee which signals the transition to \texttt{ready}.

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
+    The GameController will activate this state before \texttt{initial} (i.e at the beginning of a half and during a timeout).
+
+    \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
     Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
+    The robots are preparing to receive a visual signal from the referee.  
 
   \item[Ready.] In this state, the robots walk to their legal positions for kick-off (\cf \cref{sec:kick-off}) or a penalty kick (\cf \cref{sec:penalty_kick}).
     They remain in this state, until the head referee decides that there is no significant progress, up to a maximum of \qty{\KickOffAutoTime}{\second} for a kick-off and \qty{\PenaltyKickSetupTime}{\second} for a penalty kick.
     The GameController can activate sub-states for kick-off and penalty kicks.
+    The robot can only enter this state by either receiving a message from the GameController or detecting the visual signal or by receiving a message from a teammates who did.
```
rvilling, line 47: Since there are 3 possibilities, delete "either"

rules/game_process.tex
```diff
@@ -75,6 +81,14 @@ \subsection{Robot States}
   \label{fig:robot_states}
 \end{figure}
 
+At the scheduled time, the referee will announce that the transition from the \texttt{setup} to \texttt{initial} will be executed 60 seconds later.
+The referee shall call ``Initial in 60 seconds''. 
+The GameController will execute this transition automatically once the time has passed. 
+If both teams have left the field, the head referee can announce the transition earlier by calling ``Initial''.
```
rvilling, line 87: This sounds odd. I guess you are referring to the human team members and that the robot team members are on the field. Rewording needed I think.

rules/game_process.tex
```diff
@@ -75,6 +81,14 @@ \subsection{Robot States}
...
+
+
+The referee should wait a random time between 10 to 120 seconds before announcing the transition to \texttt{ready}. 
+Teams should expect a different time for every time they have to effectuate the transition from the \texttt{initial} to \texttt{ready}.
 The referee will announce the transition from the \texttt{initial} to \texttt{ready} state by raising both hands over its head. The referee will wear red gloves \footnote{Teams are encouraged to not rely on the red gloves as they could be removed in a following year.}.
```
rvilling, line 92: "by raising both hands over its head" --> "by raising both hands over their head"

rules/game_process.tex
arnemoos: Yes, why not go into setup and then a second chest button press transitions to the penalized state (so no real change for the handling except the first reachable state via button interface is setup)

rules/forbidden_actions.tex
```diff
@@ -110,9 +110,9 @@ \subsection{Forbidden Motion}
 \subsubsection{Motion in Initial}
 \label{sec:motion_in_initial}
 
-Robots that begin moving before the transition from \texttt{initial} to \texttt{ready} (\ie before the head referee does the visual signal).
+Robots that begin moving during \texttt{initial} (\ie before the head referee does the visual signal).
```
arnemoos, line 113: begin to move

rules/game_process.tex
rvilling: I believe that the text still assumes that the transition to calibration takes place from initial (which can only be reached via GC). Truthfully, I would have dealt with all this differently by leaving the initial state the way it originally was and inserting a new state between initial and ready to accept the ref signal - this would have resulted in fewest changes to robot code. If you want to continue in your current direction with the setup state, then I think you will also need to enable a transition from the setup state to calibration rather than from initial to calibration.

ben47955: Having the new state before or after initial is just a matter of label as at the end the definition will be the same. To my opinion we should keep the current direction. At least I will update with this direction until we decided otherwise.

ben47955:
Change are ready to by reviewed

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
     During any state, if all head buttons are pressed at least one second, the robot should move to a safe seated/crouched position and unstiffen all joints.
     So while in the \texttt{unstiff} state the robot is not allowed to move in any fashion! After booting, the robots are in their \texttt{unstiff} state.
-    Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the \texttt{initial} state, or a state as indicated by GameController.
+    Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the \texttt{setup} state, or a state as indicated by GameController.
 
-  \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
+    \item[Setup.] The robots are free to move at teams convenience and human are allowed to interact with the robots. 
+    This state is not limited in time and teams have access to the field.
+    The GameController will activate this state before \texttt{initial} (i.e at the beginning of a half and during a timeout).
```
sseering, line 38: I think we should have a setup phase only before every half and not at the end of timeout. This was also discussed when we talked about GC implementation.

Everything a team could do during setup they can do during timeout. So we don't need a setup at the end of a timeout.

rules/game_process.tex
```diff
@@ -75,7 +81,16 @@ \subsection{Robot States}
   \label{fig:robot_states}
 \end{figure}
 
-The referee will announce the transition from the \texttt{initial} to \texttt{ready} state by raising both hands over its head. The referee will wear red gloves \footnote{Teams are encouraged to not rely on the red gloves as they could be removed in a following year.}.
+At the scheduled time, the referee will announce that the transition from the \texttt{setup} to \texttt{initial} will be executed 60 seconds later.
+The referee shall call ``Initial in 60 seconds''. 
+The GameController will execute this transition automatically once the time has passed.
```
sseering, line 86: I don't think we should introduce a GC countdown here. There might be discussion between teams and the headref or GC operator. 

The transition from setup to inital simply happens when the GC decides.

rules/game_process.tex
```diff
@@ -75,7 +81,16 @@ \subsection{Robot States}
...
 \end{figure}
 
-The referee will announce the transition from the \texttt{initial} to \texttt{ready} state by raising both hands over its head. The referee will wear red gloves \footnote{Teams are encouraged to not rely on the red gloves as they could be removed in a following year.}.
+At the scheduled time, the referee will announce that the transition from the \texttt{setup} to \texttt{initial} will be executed 60 seconds later.
+The referee shall call ``Initial in 60 seconds''. 
+The GameController will execute this transition automatically once the time has passed. 
+If all humans from playing teams have left the field, the head referee can announce the transition earlier by calling ``Initial''.
+
+
+The referee should wait a random time between 10 to 120 seconds before announcing the transition to \texttt{ready}.
```
rvilling, line 90: Since the referee must not signal until the initial state is active, I suggest:

"Once the game controller operator has confirmed the initial state, the referee should wait a random time..."

rules/game_process.tex
```diff
@@ -75,7 +81,16 @@ \subsection{Robot States}
...
 
-The referee will announce the transition from the \texttt{initial} to \texttt{ready} state by raising both hands over its head. The referee will wear red gloves \footnote{Teams are encouraged to not rely on the red gloves as they could be removed in a following year.}.
+At the scheduled time, the referee will announce that the transition from the \texttt{setup} to \texttt{initial} will be executed 60 seconds later.
+The referee shall call ``Initial in 60 seconds''. 
+The GameController will execute this transition automatically once the time has passed. 
+If all humans from playing teams have left the field, the head referee can announce the transition earlier by calling ``Initial''.
+
+
+The referee should wait a random time between 10 to 120 seconds before announcing the transition to \texttt{ready}. 
+Teams should expect a different time for every time they have to effectuate the transition from the \texttt{initial} to \texttt{ready}.
```
rvilling, line 91: "Teams should expect that the delay before the referee signals the transition from  \texttt{initial} to \texttt{ready} will be different every time."

rules/game_process.tex
```diff
@@ -75,7 +81,16 @@ \subsection{Robot States}
...
+The referee will announce the transition from \texttt{initial} to \texttt{ready} state by raising both hands over their head.
+The referee will wear red gloves \footnote{Teams are encouraged to not rely on the red gloves as they could be removed in a following year.}.
 The referee must stand on the opposite side of the technical area behind the touchline and at the height of the halfway line.
```
rvilling, line 94: "at the height of" --> "in line with"

rules/game_process.tex
rvilling: I agree with Stefan - let's go with one option. I vote for Cyan. I think it will be helpful to teams to distinguish the states.

ahasselbring:
The GameController knows the setup state now (when built from the current master). Maybe you can check whether it is compatible with your intentions?

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
+    Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the \texttt{setup} state, or a state as indicated by GameController.
 
-  \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
+    \item[Setup.] The robots are free to move at teams convenience and human are allowed to interact with the robots. 
+    This state is not limited in time and teams have access to the field.
+    The GameController will activate this state before \texttt{initial} (i.e at the beginning of a half and during a timeout).
     Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
 
+    \item[Initial.] The robots are not allowed to move in any fashion besides initially standing up.
+    In \texttt{iniitial}, the robots are awaiting a visual indication from the referee which signals the transition to \texttt{ready}.
```
arnemoos, line 42: initial

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
   \item[Unstiff.] It helps to facilitate a consistent and safe handling of the robots for remote competition.
     During any state, if all head buttons are pressed at least one second, the robot should move to a safe seated/crouched position and unstiffen all joints.
     So while in the \texttt{unstiff} state the robot is not allowed to move in any fashion! After booting, the robots are in their \texttt{unstiff} state.
-    Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the \texttt{initial} state, or a state as indicated by GameController.
+    Pressing the chest button once while in the \texttt{unstiff} state, permits the robot to stiffen its joints and return to the \texttt{setup} state, or a state as indicated by GameController.
 
-  \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
+    \item[Setup.] The robots are free to move at teams convenience and human are allowed to interact with the robots.
```
arnemoos, line 36: humans?

rules/game_process.tex
arnemoos: Which state is then active during timeout?

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
+    The GameController will activate this state before \texttt{initial} (i.e at the beginning of a half and during a timeout).
     Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
 
+    \item[Initial.] The robots are not allowed to move in any fashion besides initially standing up.
+    In \texttt{iniitial}, the robots are awaiting a visual indication from the referee which signals the transition to \texttt{ready}.  
+
   \item[Ready.] In this state, the robots walk to their legal positions for kick-off (\cf \cref{sec:kick-off}) or a penalty kick (\cf \cref{sec:penalty_kick}).
     They remain in this state, until the head referee decides that there is no significant progress, up to a maximum of \qty{\KickOffAutoTime}{\second} for a kick-off and \qty{\PenaltyKickSetupTime}{\second} for a penalty kick.
     The GameController can activate sub-states for kick-off and penalty kicks.
+    The robot can only enter this state by receiving a message from the GameController or detecting the visual signal or by receiving a message from a teammates who did.
```
arnemoos, line 47: from teammates

rules/game_process.tex
ben47955: In the GC "timeout" is a phase. Having the robot in "setup" don't contradict that. Plus, it add consistency to have a setup before every initial. But I'm open to discuss that.

rules/game_process.tex
ben47955: This was your suggestion in another comment unless I misunderstand. I liked that because it would help to have matchs starting in time and make more clear the time window teams are having to call a timeout. 

What we can do is keeping the countdown but the transition to be done manually by the GC operator.

rules/game_process.tex
ahasselbring: > In the GC "timeout" is a phase.

In the message that is sent to the robots, yes. But in the GameController's internal logic, it is a state on the same level as Initial/Ready/Set/.... This is because the combination Penalty Shoot-out + Timeout is possible, but up to now, combinations of Timeout with any other state than Initial weren't possible.

rules/game_process.tex
ahasselbring: Does the suggestion of the countdown mean that the GameController should know when the match starts?

ben47955: I was thinking to rely on the referee to start on time. It would add so many issues if let the GC starting match without manual input.

rules/game_process.tex
ben47955: What if we keep Timeout on GC side, but send Setup to robot ? Robot should recover the same way from a timeout than the beginning of a half. Any opinions?

rules/game_process.tex
ahasselbring: OK, I was reading this as if the countdown (even without automatic transition, which I consider a bad idea, too) should be displayed by the GC. If it's just the referee looking at a clock and announcing the time, it's fine.

rules/game_process.tex
ahasselbring: You would still have to switch to sending Initial to the robots so they look for the referee again. If the Initial state starts only once the timeout is over, games would be delayed even more.

ben47955: I'm not sure how this would add more delay ? Robot should be ready to play before the end of the Setup. ( There the edge case of prematurely called end of the timeout as it would require human input anyway I would say it's fine)

rules/game_process.tex
ben47955: Just to be clear I mean the operator to start the initial countdown manually ( 60s before the transition to initial). The current discussion tends to also trigger the transition manually which I will update. 
There already a 1 minutes rules and 2 minutes rules about initial kick-off. This is just a matter of better defining when this start (Section 3.5). For a matter of clarity the GC should display this countdown.

rules/game_process.tex
ahasselbring: If the transition to Initial happens after full 5 minutes and then there are up to 2 minutes of Initial on top, this is longer than before.

Anyway, I have modified the GameController so that a Timeout is equivalent rather to Setup than to Initial, but the Timeout timer keeps running after the switch to Initial. The timer is not affected by this. The penalty shoot-out also starts with setup now.

There are still a couple of edge cases around request for pick-up to resolve for me, especially when you can pick-up a player without getting a timer, and how the timer should be set when the player already has a Forbidden Motion in Initial penalty (this is currently exploitable).

ben47955: This 2 minutes is a different discussions. In fact, it's between 10 sec and 2 minutes. This is the drawback to have a more predictable signal. This delay only start when robots are ready to position. I'm still not sure I understand what we call/do before Initial affect this. 
Plus, why my last commit, initial can start sooner to save couple seconds.

rules/game_process.tex
```diff
@@ -31,14 +31,20 @@ \subsection{Robot States}
...
+    The GameController will activate this state before \texttt{initial} (i.e at the beginning of a half and during a timeout).
     Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
 
+    \item[Initial.] The robots are not allowed to move in any fashion besides initially standing up.
+    In \texttt{initial}, the robots are awaiting a visual indication from the referee which signals the transition to \texttt{ready}.  
+
   \item[Ready.] In this state, the robots walk to their legal positions for kick-off (\cf \cref{sec:kick-off}) or a penalty kick (\cf \cref{sec:penalty_kick}).
     They remain in this state, until the head referee decides that there is no significant progress, up to a maximum of \qty{\KickOffAutoTime}{\second} for a kick-off and \qty{\PenaltyKickSetupTime}{\second} for a penalty kick.
     The GameController can activate sub-states for kick-off and penalty kicks.
+    The robot can only enter this state by receiving a message from the GameController or detecting the visual signal or by receiving a message from teammates who did.
```
ahasselbring, line 47: Is this sentence really required? It's missing the whistle for the transition after a goal.

ben47955: Yes as there nothing else preventing educated guess of the signal without visual recognition. I will add the whistle to the list or see if I can do otherwise.

rules/game_process.tex
```diff
@@ -33,12 +33,24 @@ \subsection{Robot States}
...
+    The GameController will activate this state before \texttt{visual} (i.e at the beginning of a half and during a timeout).
     Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
 
+    \item[Visual.] The robots are not allowed to move in any fashion besides initially standing up and moving their heads.
+    In \texttt{visual}, the robots are awaiting a visual indication from the referee which signals the transition to \texttt{ready}.  
+
   \item[Ready.] In this state, the robots walk to their legal positions for kick-off (\cf \cref{sec:kick-off}) or a penalty kick (\cf \cref{sec:penalty_kick}).
     They remain in this state, until the head referee decides that there is no significant progress, up to a maximum of \qty{\KickOffAutoTime}{\second} for a kick-off and \qty{\PenaltyKickSetupTime}{\second} for a penalty kick.
     The GameController can activate sub-states for kick-off and penalty kicks.
+    The robot can only enter this state by :
```
ahasselbring, line 47: I'm still not convinced of this enumeration. Estimating the state of the game can integrate a lot of cues. For instance, why should it be illegal to enter the ready state when you (visually) observe other players entering the pitch, or (after a goal) when the ball is observed to have moved in a goal and subsequently removed? Humans certainly do this to disambiguate the state of the game and don't rely only on a sound or a signal.

I agree that there are "outside-the-box" things (such as comparing the robot's clock with the scheduled start of the match, or waiting a fixed time and such) that shouldn't be done. But my guess is that those are bad ideas that don't work well anyway. And otherwise, this is just a reiteration of the general "please don't cheat" principle.

ben47955: I removed the list. I was hesitating to move the meaning as exit condition of Visual state, but with all the instruction about the visual signal, let say strategy not involving the visual signal at all would be against the spirits of the game.

rules/forbidden_actions.tex
```diff
@@ -107,12 +107,12 @@ \subsection{Forbidden Motion}
 ``Forbidden Motion'' penalties do not follow the standard removal procedure, and hence do not count towards the incremental penalty count.
 A robot will be moved back to its original position if it has moved significantly before becoming penalized.
 
-\subsubsection{Motion in Initial}
-\label{sec:motion_in_initial}
+\subsubsection{Motion in Visual}
```
arnemoos, line 110: I believe the current name "Visual" for the state where robots are waiting for the referee's gesture is not very intuitive. I suggest we use **"Standby"** instead, as it clearly conveys a state of readiness while awaiting further instructions.

rules/game_process.tex
```diff
@@ -33,9 +33,15 @@ \subsection{Robot States}
...
 
-  \item[Initial.] The robots are not allowed to be moving in any fashion besides initially standing up.
+    \item[Initial.] The robots are free to move at teams convenience and humans are allowed to interact with the robots. 
+    This state is not limited in time and teams have access to the field.
+    The GameController will activate this state before \texttt{visual} (i.e at the beginning of a half and during a timeout).
     Shortly pressing the chest button will switch the robot to the \texttt{penalized} state.
 
+    \item[Visual.] The robots are not allowed to move in any fashion besides initially standing up and moving their heads.
+    In \texttt{visual}, the robots are awaiting a visual indication from the referee which signals the transition to \texttt{ready}. 
+    This state is used for Champions Cup matches only. Reference to \texttt{visual} in a Challenge Shield matches refer to \texttt{ready}.
```
arnemoos, line 43: I would rephrase the second sentence to be more like: 

This state is used for Champions Cup matches only. For Challenge Shield matches, this state is skipped, and it is transitioned directly to \texttt{ready}.

rules/game_process.tex
```diff
@@ -75,10 +81,20 @@ \subsection{Robot States}
   \label{fig:robot_states}
 \end{figure}
 
-The referee will announce the transition from the \texttt{initial} to \texttt{ready} state by raising both hands over its head. The referee will wear red gloves \footnote{Teams are encouraged to not rely on the red gloves as they could be removed in a following year.}.
-The referee must stand on the opposite side of the technical area behind the touchline and at the height of the halfway line.
+At the scheduled time, the referee will announce that the transition from the \texttt{initial} to \texttt{visual} will be executed 60 seconds later.
+The referee shall call ``Start in 60 seconds''
+The GameController operator shall start the countdown through the GameController.
```
arnemoos, line 86: I am not sure if the GameController needs a countdown timer?

rules/game_process.tex
```diff
@@ -75,10 +81,20 @@ \subsection{Robot States}
...
+The GameController operator will announce the transition from \texttt{initial} to \texttt{visual} when the head referee is calling ``Start''.
```
arnemoos, line 87: Here we normally just call the current start and I wouldn't call "Start" here since this assumes that the game really starts and then nothing happens for a random amount of time. So here it would also come in handy if the state is called Standby and the referee could just announce Standby.

rules/game_process.tex
```diff
@@ -75,10 +81,20 @@ \subsection{Robot States}
...
+If all humans from playing teams have left the field, the head referee can announce the transition earlier.
+
+
+Once the game controller operator has confirmed the \texttt{visual} state, the referee should wait a random time between 10 to 120 seconds before announcing the transition to \texttt{ready}. 
+Teams should expect that the delay before the referee signals the transition from \texttt{visual} to \texttt{ready} will be different every time.
+The referee will announce the transition from \texttt{visual} to \texttt{ready} state by raising both hands over their head.
+The referee will wear red gloves \footnote{Teams are encouraged to not rely on the red gloves as they could be removed in a following year.}.
+The referee must stand on the opposite side of the technical area behind the touchline and in line with the halfway line.
```
arnemoos, line 95: Three sentences beginning with "the". Maybe better:

To announce the transition from \texttt{visual} to \texttt{ready} state, the referee will raise both hands over their head.
Red gloves will be worn by the referee, though teams are encouraged not to rely on this visual cue as the gloves could be removed in a following year. 
The referee must stand on the opposite side of the technical area, behind the touchline, and in line with the halfway line.

arnemoos (changes requested):

rules/game_process.tex
ahasselbring: At least I don't understand this enough to implement it and would rather not do it (it seems an unnecessary extra step to me).

ben47955: The idea is to add consistency between referee and avoiding unfairness of having one teams stiffen while the other team still deploying his code. 
Also, one challenge we discussed about removing some state and having robot allowed to move is it's harder to know when exactly the game start. Making the start time rely less on the referee. 

But I will remove the GC part.

rules/forbidden_actions.tex
ben47955: I will make the update

ahasselbring:
Does the TC have an opinion whether it should be possible to take a timeout in the new state?

ben47955:
> Does the TC have an opinion whether it should be possible to take a timeout in the new state?

I don't really see how we can forbid them while still allowing them in Ready. But we will have a word on that.

ben47955:
Name of the state updated again along others comments

ahasselbring:
> > Does the TC have an opinion whether it should be possible to take a timeout in the new state?
> 
> I don't really see how we can forbid them while still allowing them in Ready. But we will have a word on that.

That makes sense.

arnemoos (approved):

PR #158, opened 2024-07-01
Attempt use 90s and adding challenge points

ben47955:


ahasselbring:
What is the reasoning behind those changes at this point?
Especially, continuing the challenge after a goal seems quite a big change to me (and is not supported by the GameController at the moment).

jpha226:
Hi Arne,

The concern is that only a single goal might not differentiate teams well enough. For example, one team might score in 15 seconds and another takes the full 90 seconds. The first team would appear to be stronger but each team would have an equal result (ignoring what happens on defense). This could be a problem for tie-breakers since each team only gets to play a fraction of the total teams.

Is this a big or small change to the GameController? I thought this would be already supported or at most a small change. If it is a bigger change, we can discuss an alternative way to address the above issue (e.g., using time-to-score as a tie-breaker).

Josiah

From: Arne Hasselbring ***@***.***>
Date: Tuesday, July 2, 2024 at 6:47 AM
Cc: Subscribed ***@***.***>
Subject: Re: [RoboCup-SPL/Rules] Attempt use 90s and adding challenge points (PR #158)

What is the reasoning behind those changes at this point?
Especially, continuing the challenge after a goal seems quite a big change to me (and is not supported by the GameController at the moment).

roefer:
I don't think that the rules should be changed two weeks before the challenge takes place. The original rules were designed so that teams should perform passes rather than scoring goals, because that's how you can achieve many points. Changing the main objective of the challenge so short before the competition is unfair for the teams who already prepared for the challenge.

jpha226:
Closing pull request as we determined not to change anything.

PR #162, opened 2024-10-29
New visual signal for set plays + indirect kick change for RoboCup 2025

torchipeppo:
Left some red text and todo-notes for the greatest points of discussion.

### DIFF
--- a/rules/app_competition_rules.tex
+++ b/rules/app_competition_rules.tex
@@ -60,7 +60,7 @@ Matches in the Champions Cup are played with up to 7 players on each team, while
 In addition, some rules are simplified for Challenge Shield matches. The major differences are:
 \begin{itemize}
   \item The requirement to recognize visual referee signals is relaxed for the Challenge Shield (\cf~\cref{sec:robot_players}).
-  \item Goals may be scored via direct kick from kick-off. The indirect kick rules remain unchanged for the Challenge Shield (\cf~\cref{sec:indirect_kick_old}), but were tightened for the Champions Cup.
+  \item Although the indirect kick rules are the same for both competitions, Challenge Shield teams have the option to have them relaxed if they declare they cannot play with more than \MaxRobotsInFallback robots. (See \cref{sec:fallback_mode})
 \end{itemize}
 
 \subsection{Game Structure}
--- a/rules/app_competition_rules.tex
+++ b/rules/app_competition_rules.tex
@@ -60,7 +60,7 @@ Matches in the Champions Cup are played with up to 7 players on each team, while
 In addition, some rules are simplified for Challenge Shield matches. The major differences are:
 \begin{itemize}
   \item The requirement to recognize visual referee signals is relaxed for the Challenge Shield (\cf~\cref{sec:robot_players}).
-  \item Goals may be scored via direct kick from kick-off. The indirect kick rules remain unchanged for the Challenge Shield (\cf~\cref{sec:indirect_kick_challenger}), but were tightened for the Champions Cup.
+  \item Goals may be scored via direct kick from kick-off. The indirect kick rules remain unchanged for the Challenge Shield (\cf~\cref{sec:indirect_kick_old}), but were tightened for the Champions Cup.
 \end{itemize}
 
 \subsection{Game Structure}
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -165,11 +165,10 @@ There are two exceptions to this rule:
 \begin{itemize}
   \item A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick.
   \item If a team is awarded a free kick but fails to execute it within the allotted time, the opposing team may attempt to score directly if they play the ball before the team who was originally awarded the free kick does.
-    ("Execute" is intended as "kick and move the ball clearly", as per section \cref{sec:free_kick_execution})
+    (``Execute'' is intended as "kick and move the ball clearly", as per section \cref{sec:free_kick_execution})
 \end{itemize}
 
 In both cases, the direct goal is only allowed if the kicking robot played the ball once.
-\unsure{Updated after discussion, but not closed yet}
 The indirect kick rule returns into effect when the ball comes to a full stop after the kick, or if the same robot plays the ball again before then.
 In this case, any contacts made by the kicking robot, or any other robot, during the direct kick count towards the indirect kick rule.
 
@@ -226,9 +225,9 @@ the number of robots the opposing team is allowed to play with.
 
 \info[inline]{Memo for the GC implementation: the excess robots are \textbf{not}
 picked up, they are ``outside the game'', or ``disabled'', or w/e.
-Substitution via pick-up should transfer the 45-second timer as usual.}
+Substitution via pick-up should transfer the pickup/penalty timer as usual.}
 
-Teams decide which operational mode to play in before the beginning of each half.
+Teams decide which operational mode they play in before the beginning of each half.
 Furthermore, during any timeout (regardless of who initiated it), each team
 may choose to switch from \textit{normal mode} to \textit{fallback mode}, but not vice versa.
 In both cases, teams must inform the head referee and the GameController operator
@@ -461,8 +460,8 @@ The head referee is encouraged to seek the acknowledgement of the GameController
 
 The referee makes an effort to send the signal standing as close as possible to the T-junction joining the half-way line to the touchline opposite to the technical area.
 However, their actual positioning along the touchline may deviate slightly from this standard position, based on the current game situation and the timing of the gesture.
-\unsure{tentative} Teams are not required to scan the entire touchline:
-in broad terms, a robot standing about \textcolor{red}{\textbf{(three? four?)}} meters away from the T-junction should be able to find the referee in its camera field with no or minimal head motion.
+Teams are not required to scan the entire touchline:
+in broad terms, a robot standing a few meters away from the T-junction should be able to find the referee in its camera field with no or minimal head motion.
 \info[inline]{Remind the refs that the GC operator will take a few moments to click the button, giving them a little time to make a couple steps if needed.}
 
 Teams are encouraged to make an effort to familiarize with looking at the referee and interpreting the visual signal even in cases where such an action may not seem strictly necessary,
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -165,7 +165,7 @@ There are two exceptions to this rule:
 \begin{itemize}
   \item A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick.
   \item If a team is awarded a free kick but fails to execute it within the allotted time, the opposing team may attempt to score directly if they play the ball before the team who was originally awarded the free kick does.
-    (``Execute'' is intended as "kick and move the ball clearly", as per section \cref{sec:free_kick_execution})
+    (``Execute'' is intended as ``kick and move the ball clearly'', as per section \cref{sec:free_kick_execution})
 \end{itemize}
 
 In both cases, the direct goal is only allowed if the kicking robot played the ball once.
@@ -452,7 +452,7 @@ The raised arm is the one pointing towards the defending team's goal, i.e. the g
   Remove the footnote after RC2025 has passed
 }
 \footnote{For reference, teams who participated in the 2023 Visual Referee Challenge
-  may notice that this the same as the "Kick-in" gesture from back then,
+  may notice that this the same as the ``Kick-in'' gesture from back then,
   but with the opposite arm raised.
 }
 
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -167,10 +167,10 @@ There are two exceptions to this rule:
   \item If a team is awarded a free kick but fails to execute it within the allotted time, the other team may score directly as long as the second team plays the ball first.
 \end{itemize}
 
-In both cases, the direct goal is only allowed if the kicking robot only touched the ball once.
-\unsure{Discuss what to do with accidental deflections}
-If any robot (whether the same or a different one) plays the ball again, then the indirect kick rule returns into effect.
-In this case, the contact made by the kicking robot counts towards the indirect kick rule.
+In both cases, the direct goal is only allowed if the kicking robot played the ball once.
+\unsure{Updated after discussion, but not closed yet}
+The indirect kick rule returns into effect when the ball comes to a full stop after the kick, or if the same robot plays the ball again before then.
+In this case, any contacts made by the kicking robot, or any other robot, during the direct kick count towards the indirect kick rule.
 
 \begin{description}
   \item[Example 1:] Player 2 (of the red team) kicks the ball to Player~3 (of the red team), who then kicks the ball into the blue team's goal.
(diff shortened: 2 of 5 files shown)


rules/game_process.tex
```diff
@@ -340,6 +364,17 @@ \subsection{Free Kick}
...
 The team who is awarded the free kick (termed the attacking team) has \qty{\FreeKickTime}{\second} to complete the kick.
 
+For most of the free kicks in Champions Cup, the referee communicates which team is to take the free kick via a visual gesture.
+In this case, the GameController packet will tell the robots which free kick sub-state is active, but not who is to kick the ball.
+\unsure{This gesture is entirely tentative and just my (Francesco's) proposal, pending discussion with the TC. The intent is to mirror the FIFA gestures in a simplified way.}
+The gesture consists of raising one arm horizontally and sideways, parallel to the field line, while the other arm remains at rest.
+The raised arm is the one pointing towards the defending team's goal, i.e. the goal of the team that is \textit{not} taking the free kick.
+This gesture is made from \textcolor{red}{\textbf{[a specified position, or anywhere along the field border]}} and held for \textcolor{red}{\textbf{[no less than 2 seconds, and possibly more]}}.
+
+The above rule applies to the following free kicks: \textcolor{red}{\textbf{[list of set plays this applies to]}}.
```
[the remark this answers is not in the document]
torchipeppo: I'm mostly wondering about the penalty kick, since it has the additional complexity of the ready phase, but it's not that crippling, just something to think about and confirm before finalizing.

rules/game_process.tex
```diff
@@ -340,6 +364,19 @@ \subsection{Free Kick}
...
+\unsure{This gesture is entirely tentative and just my (Francesco's) proposal, pending discussion with the TC. The intent is to mirror the FIFA gestures in a simplified way.}
+The gesture consists of raising one arm horizontally and sideways, parallel to the field line, while the other arm remains at rest.
+The raised arm is the one pointing towards the defending team's goal, i.e. the goal of the team that is \textit{not} taking the free kick.
+\textit{(For reference, the gesture is the same as the "Kick-in" gesture from the 2023 Visual Referee Challenge, but the opposite arm is raised with respect to back then.)}
+This gesture is made by the head referee and held for at least 5 seconds after the GameController signal is sent.
+\textit{(The referee may look at the GameStateVisualizer monitor to check that the GC signal is sent and when the time allotted for the gesture has passed.)}
+The referee sends the signal standing as close as possible to the T-junction, but their actual positioning along the sideline may deviate from this standard position, even significantly,
```
torchipeppo, line 375: Discuss: might this mean that the head ref won't be able to stand behind the goal (physical structure) to adjudicate close goals (points), where the ball barely crosses the line? Can/Should an assistant do this?

rules/game_process.tex
```diff
@@ -158,13 +158,24 @@ \subsubsection{Indirect Kick - Champions Cup}
     A deliberate play at the ball includes successfully kicking the ball, dribbling the ball or the goalkeeper playing at the ball with its hands.
   \item If an indirect free kick is kicked directly into the opponent's goal, the goal does not count and a \texttt{Goal kick} is awarded.
   \item Note that an own-goal may always be scored without requiring an indirect kick.
-  \item A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick. A contact in the penalty kick counts among this rule.
 \end{itemize}
 
+There are two exceptions to this rule:
+
+\begin{itemize}
+  \item A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick.
```
arnemoos, line 166: A contact in the penalty kick counts among this rule.

rules/game_process.tex
```diff
@@ -158,13 +158,24 @@ \subsubsection{Indirect Kick - Champions Cup}
...
+  \item If a team is awarded a free kick but fails to execute it within the allotted time, the other team may score directly as long as the second team plays the ball first.
```
arnemoos, line 167: unsure if second team is also "the other team"?

rules/game_process.tex
arnemoos: I would say there should be mentioned something that the head referee should try to reach the T-Junction after calling a set play. For a goal its not a problem since no set play follows but for a goalkick or corner kick this could happen that the ref deviated a bit from the T-Junction. 

We also had the idea that there might be a 5th referee involved.

rules/game_process.tex
```diff
@@ -340,6 +364,19 @@ \subsection{Free Kick}
 Note that in the case of the Pushing free kick the sub-state is activated automatically through the ``Foul''.
 The team who is awarded the free kick (termed the attacking team) has \qty{\FreeKickTime}{\second} to complete the kick.
 
+In Champions Cup, the referee communicates which team is to take the free kick via a visual gesture.
+The GameController packet will tell the robots which free kick sub-state is active, but not who is to kick the ball.
+\unsure{This gesture is entirely tentative and just my (Francesco's) proposal, pending discussion with the TC. The intent is to mirror the FIFA gestures in a simplified way.}
+The gesture consists of raising one arm horizontally and sideways, parallel to the field line, while the other arm remains at rest.
+The raised arm is the one pointing towards the defending team's goal, i.e. the goal of the team that is \textit{not} taking the free kick.
+\textit{(For reference, the gesture is the same as the "Kick-in" gesture from the 2023 Visual Referee Challenge, but the opposite arm is raised with respect to back then.)}
+This gesture is made by the head referee and held for at least 5 seconds after the GameController signal is sent.
```
arnemoos, line 373: Maybe add something like after the GC controller acknowledged it? Or the transition is indicated by the GameStateVisualizer.

rules/game_process.tex
```diff
@@ -340,6 +364,19 @@ \subsection{Free Kick}
...
+\textit{(The referee may look at the GameStateVisualizer monitor to check that the GC signal is sent and when the time allotted for the gesture has passed.)}
+The referee sends the signal standing as close as possible to the T-junction, but their actual positioning along the sideline may deviate from this standard position, even significantly,
+e.g. if the referee needs to better observe a specific situation during the game.
+
+In Challenge Shield, this visual signal is not performed and the kicking team is communicated by the GameController.
```
arnemoos, line 378: Last year with the indirect kick there was a subsection for champions cup and one for challenge shield and this could be handled this year in the same way?

arnemoos (changes requested):

rules/game_process.tex
torchipeppo: If you mean that the second sentence in line 161- should be replicated here, I'll respond immediately because I already know the answer: that concept is expressed in line 173+, with the intention that the rules after this bullet list apply to both types of direct kick.

rules/game_process.tex
torchipeppo: Oh, nice ambiguity spotting skills. I was trying to describe the situation without making up names such as "Team A" and "Team B", but I already had a little difficulty.

ben47955:
Something that we discussed over call, but didn't over our call is to rename the goal kick and corner kick into the same name send in the packet to blur the line more.

rules/game_process.tex
```diff
@@ -158,13 +158,25 @@ \subsubsection{Indirect Kick - Champions Cup}
...
   \item If an indirect free kick is kicked directly into the opponent's goal, the goal does not count and a \texttt{Goal kick} is awarded.
   \item Note that an own-goal may always be scored without requiring an indirect kick.
-  \item A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick. A contact in the penalty kick counts among this rule.
 \end{itemize}
 
+There are two exceptions to this rule:
+
+\begin{itemize}
+  \item A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick.
+  \item If a team is awarded a free kick but fails to execute it within the allotted time, the other team may score directly as long as the latter team plays the ball first.
```
rvilling, line 167: Confusing and likely incorrect to say "other team" and then "latter team".  Where you've used latter team are you referring to the team originally awarded the penalty who timed out?

rules/game_process.tex
```diff
@@ -340,6 +365,21 @@ \subsection{Free Kick}
 Note that in the case of the Pushing free kick the sub-state is activated automatically through the ``Foul''.
 The team who is awarded the free kick (termed the attacking team) has \qty{\FreeKickTime}{\second} to complete the kick.
 
+In Champions Cup, the referee communicates which team is to take the free kick via a visual gesture.
+The GameController packet will tell the robots which free kick sub-state is active, but not who is to kick the ball.
+\unsure{This gesture is entirely tentative and just my (Francesco's) proposal, pending discussion with the TC. The intent is to mirror the FIFA gestures in a simplified way.}
+The gesture consists of raising one arm horizontally and sideways, parallel to the field line, while the other arm remains at rest.
+The raised arm is the one pointing towards the defending team's goal, i.e. the goal of the team that is \textit{not} taking the free kick.
```
rvilling, line 372: It would be easier to say that the raised arm is pointing in the direction the free kick has been given (i.e. towards the goal being attacked).

rules/game_process.tex
```diff
@@ -340,6 +365,21 @@ \subsection{Free Kick}
...
+\textit{(For reference, the gesture is the same as the "Kick-in" gesture from the 2023 Visual Referee Challenge, but the opposite arm is raised with respect to back then.)}
+This gesture is made by the head referee and held for at least 5 seconds after the GameController signal is sent.
+\textit{(The referee may look at the GameStateVisualizer monitor to check that the GC signal is sent and when the time allotted for the gesture has passed.)}
+The referee makes an effort to sends the signal standing as close as possible to the T-junction.
```
rvilling, line 376: Be more specific: "the T-junction joining the half-way line to the side line".

rules/game_process.tex
torchipeppo: Well, the only difference is this one paragraph about the gesture, unlike the indirect kick where the whole rule changes, so having a CC subsection and a CS subsec might be unnecessary.
But I think I can add just one "Visual gesture" subsection that starts with a message like "This only applies to Champions Cup", so the part that differs between the two divisions is still separate from the rest. It would also make the cross-reference from the penalty kick section more precise.

rules/game_process.tex
torchipeppo: My rationale here is that I feel like referring directly to which team is or is not taking the free kick is the most technical and unambiguous way to define the rule in this case, and the more intuitive "defending team" definition is still given first, before the more explicit one.
Furthermore, especially with the indirect kick rule, the robot may perform the free kick in any direction, so saying that the arm points "in the direction the free kick has been given" might be a little odd.

torchipeppo:
Addressed may comments here, and also turned all italicized comments into inline todo notes so they're easier to remove later.
Also, thank you very much for your in-depth reviews.

ben47955:
You may want to update line 276 in game_process.tex

ben47955:
You may want to update line 276-301 : "A goal may be scored directly from a kick-off for the Challenge Shield. However, for the Champions Cup the indirect kick rule applies and the goal does \textit{not} count."

torchipeppo:
@ben47955 Decided to simply remove that line, since now both divisions play by the same indirect kick rules by default, and move the mention of a direct kick-off in the new "indirect kick variant" section I'm about to push.

rules/game_process.tex
```diff
@@ -205,31 +205,82 @@ \subsubsection{Indirect Kick - Champions Cup}
...
+they are going to play normally or play in "reduced mode" (temporary name).
+If at least one team plays in "reduced mode", the whole game follows different,
+more lenient rules for indirect kicks.
+
+\subsubsection{"Reduced Mode" rules}
+If a team plays in "reduced mode", they are only allowed to field at most two robots.
+The other robots must be removed from the field via "Request for Pickup".
+
+Teams decide which "mode" to play in before the beginning of each half.
+Furthermore, during any team's timeout or referee timeout,
```
TomJEllis, line 225:
```suggestion
Furthermore, during any timeout,
```

rules/app_competition_rules.tex
```diff
@@ -60,7 +60,7 @@ \subsection{Champions Cup and Challenge Shield}
 In addition, some rules are simplified for Challenge Shield matches. The major differences are:
 \begin{itemize}
   \item The requirement to recognize visual referee signals is relaxed for the Challenge Shield (\cf~\cref{sec:robot_players}).
-  \item Goals may be scored via direct kick from kick-off. The indirect kick rules remain unchanged for the Challenge Shield (\cf~\cref{sec:indirect_kick_challenger}), but were tightened for the Champions Cup.
+  \item Goals may be scored via direct kick from kick-off. The indirect kick rules remain unchanged for the Challenge Shield (\cf~\cref{sec:indirect_kick_old}), but were tightened for the Champions Cup.
```
ben47955, line 63: This need to be updated

rules/game_process.tex
```diff
@@ -178,33 +190,99 @@ \subsubsection{Indirect Kick - Champions Cup}
...
+  Player 2 (of the red team) then kicks the ball and scores.
+  The exception does \textit{not} apply for the red team because a blue-team robot played the ball first. The goal does not count and a \texttt{Goal kick} is awarded to the blue team.
 \end{description}
 
-\subsubsection{Indirect Kick - Challenge Shield}
-\label{sec:indirect_kick_challenger}
+\subsubsection{Fallback mode for Challenge Shield}
+
+This rule aims to allow \textbf{Challenge Shield} teams to play games
+even in the unfortunate case where severe circumstances make them unable to
```
ben47955, line 211: We should specify early the exact number of robot early in the text description for this to apply

rules/game_process.tex
```diff
@@ -178,33 +190,99 @@ \subsubsection{Indirect Kick - Champions Cup}
...
+field enough robots to play by the default indirect kick rules.
+
+Each Challenge Shield team participating in a game shall declare their operational
+capacity for that game by choosing to play either in \textbf{normal mode} or in \textbf{fallback mode}.
+If at least one team plays in \textit{fallback mode}, the whole game follows different,
+more lenient rules for indirect kicks, detailed in \cref{sec:indirect_kick_old}.
+
+If a team plays in \textit{fallback mode}, they are only allowed to field at most two robots.
```
ben47955, line 219: Maybe specify that this doesn't affect the number of robot allowed for the other team ?

rules/game_process.tex
```diff
@@ -340,6 +416,53 @@ \subsection{Free Kick}
...
+}
+\footnote{For reference, teams who participated in the 2023 Visual Referee Challenge
+  may notice that this the same as the "Kick-in" gesture from back then,
+  but with the opposite arm raised.
+}
+
+This gesture is made by the head referee and held for at least 5 seconds after the GameController signal is sent.
+The head referee is encouraged to seek the acknowledgement of the GameController operator and/or look at the field-facing monitor (if available) in order to determine the timing.
+
+The referee makes an effort to send the signal standing as close as possible to the T-junction joining the half-way line to the touchline opposite to the technical area.
```
ben47955, line 453: I think we can improve the wording to say that looking at the T-junction will allow the signal to be visible in the camera field (assuming it is not too close or positioned awkwardly at the extremity of the field). However, teams are not expected to scan the entire sideline.

rules/game_process.tex
```diff
@@ -340,6 +416,53 @@ \subsection{Free Kick}
...
+However, their actual positioning along the touchline may deviate slightly from this standard position, based on the current game situation and the timing of the gesture.
+\info[inline]{Remind the refs that the GC operator will take a few moments to click the button, giving them a little time to make a couple steps if needed.}
+
+Teams are encouraged to make an effort to familiarize with looking at the referee and interpreting the visual signal even in cases where such an action may not seem strictly necessary,
+as more information may be communicated in this way rather than via GameController packets in future years.
+
+\textbf{In Challenge Shield}, this visual signal does not need to be performed and the kicking team is communicated by the GameController.
```
ben47955, line 460: I think we think to think more about this. If there no change in GC for this, it may be hard for teams to do that without loosing time to execute the Set Play.

rules/app_competition_rules.tex
torchipeppo: I'm going to push a proposal for an updated rule, but maybe we can also consider dropping this entirely

rules/game_process.tex
torchipeppo: I'm pushing an additional sentence to this effect, but I'm open to discussion to phrase it better.

rules/game_process.tex
torchipeppo: I'm taking too long to write up an answer so maybe this is better left for the next meeting.

I'll just say that I am aware of this, and CS games are normally expected to be played without set-play gestures. The "does not need to" phrasing just means that a team requesting the gestures to be performed, for whatever reason they might have, is not technically forbidden.

Also, let me clarify that the todo notes are not intended to appear in the final rules document, but rather they are intended to be communicated informally.

rules/game_process.tex
```diff
@@ -165,11 +165,10 @@ \subsection{Indirect Kick}
 \begin{itemize}
   \item A direct kick is allowed for the duration of the penalty kick, but this rule still applies after this kick.
   \item If a team is awarded a free kick but fails to execute it within the allotted time, the opposing team may attempt to score directly if they play the ball before the team who was originally awarded the free kick does.
-    ("Execute" is intended as "kick and move the ball clearly", as per section \cref{sec:free_kick_execution})
+    (``Execute'' is intended as "kick and move the ball clearly", as per section \cref{sec:free_kick_execution})
```
TomJEllis, line 168:
```suggestion
    (``Execute'' is intended as ``kick and move the ball clearly'', as per section \cref{sec:free_kick_execution})
```

torchipeppo: Fixed that and another hidden double quote, ty

arnemoos (approved):

PR #163, opened 2024-11-05
Referee Advancements + Gameflow Adjustments

TheJollyDuck:
Initial draft pull request to keep track of rule changes pertaining to the Referee Advancements and Gameflow Adjustments discussed in TC meetings.

### DIFF
--- a/rules/game_process.tex
+++ b/rules/game_process.tex
@@ -517,6 +517,16 @@ The GameController operator should execute this addition of time using the GameC
 A game will conclude once the game score shows a goal difference of 10.
 Ending the game is mandatory once a goal difference of 10 is reached.
 
+\subsection{Drop-Ball Rule}
+\label{sec:drop_ball_rule}
+
+In the rare event that there has been a referee error or a pause in game, [WIP] .
+
+\subsection{Early Stopping of Match}
+\label{sec:early_stopping}
+
+A game will conclude early if the ball comes to a complete stop outside of the penalty area, nearing the end of the game.
+
 \subsection{Penalty Kick Shoot-Out}
 \label{sec:penalty_shoot-out}
 
@@ -598,3 +608,15 @@ Teams take one additional penalty kick each, and the game decision will be made 
   \item Otherwise, the sudden death is repeated.
   \item If after 3 sudden death penalty kicks there is still no winner, the referee will toss a coin to decide the game.
 \end{itemize}
+
+\subsection{Aborting A Penalty}
+\label{sec:aborting_penalty}
+
+If a penalty was incorrectly applied to a robot and acknowledged by the referee to be incorrect, the penalty must be aborted and a signal must be sent from the GameController, with the affected robot being unable to move, with the same process used for a Standard Removal Penalty.
+
+The robot should then be removed and placed outside the field, following the same process for the Standard Removal Penalty. Once placed, the robot should be allowed to continue play immediately after being placed.
+
+\subsection{GameController Audio Output}
+\label{sec:gc_audio_output}
+
+After each decision being made by the head referee, the GameController must reflect that decision and should output the decision via audio.


TheJollyDuck:
Will need some clarification on the drop-ball rule (not particularly sure on what it entails).

rules/game_process.tex
```diff
@@ -517,6 +517,16 @@ \subsection{Mercy Rule}
 A game will conclude once the game score shows a goal difference of 10.
 Ending the game is mandatory once a goal difference of 10 is reached.
 
+\subsection{Drop-Ball Rule}
+\label{sec:drop_ball_rule}
+
+In the rare event that there has been a referee error or a pause in game, [WIP] .
```
arnemoos, line 523: I wouldn't use the word rare

rules/game_process.tex
```diff
@@ -517,6 +517,16 @@ \subsection{Mercy Rule}
...
+
+\subsection{Early Stopping of Match}
+\label{sec:early_stopping}
+
+A game will conclude early if the ball comes to a complete stop outside of the penalty area, nearing the end of the game.
```
arnemoos, line 528: As discussed in the last meeting this was the wrong understanding.

rules/game_process.tex
```diff
@@ -598,3 +608,15 @@ \subsubsection{Sudden Death Shoot-Out}
   \item Otherwise, the sudden death is repeated.
   \item If after 3 sudden death penalty kicks there is still no winner, the referee will toss a coin to decide the game.
 \end{itemize}
+
+\subsection{Aborting A Penalty}
+\label{sec:aborting_penalty}
+
+If a penalty was incorrectly applied to a robot and acknowledged by the referee to be incorrect, the penalty must be aborted and a signal must be sent from the GameController, with the affected robot being unable to move, with the same process used for a Standard Removal Penalty.
+
+The robot should then be removed and placed outside the field, following the same process for the Standard Removal Penalty. Once placed, the robot should be allowed to continue play immediately after being placed.
```
arnemoos, line 617: If a penalty was incorrectly applied to a robot and acknowledged by the referee to be incorrect, the robot should first be removed and placed outside the field, following the same process for the Standard Removal Penalty. Once placed, the robot should then be allowed to continue play immediately after being placed and the GameController removes its penalty.

rules/game_process.tex
```diff
@@ -598,3 +608,15 @@ \subsubsection{Sudden Death Shoot-Out}
...
+
+\subsection{GameController Audio Output}
+\label{sec:gc_audio_output}
+
+After each decision being made by the head referee, the GameController must reflect that decision and should output the decision via audio.
```
arnemoos, line 622: I think the GC Audio Output was not finally decided on?

arnemoos (changes requested):

TheJollyDuck:
I have updated the Ball-Stop Rule according to what I remember.

I have also fleshed out the drop ball rule, though I have some question as to how it should be approached. Following how the proper soccer rules are laid out, the ball is placed where it was last played before a disruption.

The way the robots are positioned needs to be clarified I think, or I may be missing something.

1. Would the robots be walking back to position in a Standby->Ready fashion, but it would not be a proper kick-off?
2. Is the idea of placing it at last position be something we want?

TheJollyDuck:
I have also removed the section for the GC Audio Output for now, as I thought it may be needed considering it is still on the list.

TheJollyDuck:
For the referee change, it originally had the line about the assistant referees only being allowed to indicate violations.
The change reflects the idea that this requirement should be flexible, and change according to what the head ref would like for handling a game.

rules/judgment.tex
```diff
@@ -80,7 +82,15 @@ \subsection{Referee--Team Communication}
...
+\label{sec:referee_during_match}
 
 The head referee and the assistant referees should wear socks \emph{of black or dark blue color} and avoid reserved colors (white and green) in their leg clothing.
 They may enter the field in particular situations, \eg, to remove a robot when applying a penalty.
 They should avoid interfering with the robots as much as possible.
+
+\subsection{Referee List For Friendly Games}
+\label{sec:referee_list}
+
+During the RoboCup competition setup days and during the game days, several teams may want to participate in friendly games with each other.
```
ben47955, line 94: Maybe add that OC maintain the list and to refer to them to get on the list.

TheJollyDuck:
The drop-ball rule change is done according to what was discussed previously. The only thing that is needed to verify is that the GameController does not have a kickoff timer when going straight to the *playing* state, as it would be seen as wasted time.

The temporary approach would be to go straight to ball free in this regard.

torchipeppo (changes requested):

rules/game_process.tex
```diff
@@ -598,3 +608,8 @@ \subsubsection{Sudden Death Shoot-Out}
   \item Otherwise, the sudden death is repeated.
   \item If after 3 sudden death penalty kicks there is still no winner, the referee will toss a coin to decide the game.
 \end{itemize}
+
+\subsection{Aborting A Penalty}
+\label{sec:aborting_penalty}
+
+If a penalty was incorrectly applied to a robot and acknowledged by the referee to be incorrect, the robot should first be removed and placed outside the field, following the same process for the Standard Removal Penalty. Once placed, the robot should then be allowed to continue play immediately after being placed and the GameController removes its penalty.
```
rvilling, line 615: In this case the penalty count for the team should not include the aborted penalty.
Furthermore, the penalty time increase described in the following text should not apply to an aborted penalty: "Unless otherwise specified, the penalty time increases by 10s each time a team commits any infraction"

rules/game_process.tex
```diff
@@ -517,6 +517,16 @@ \subsection{Mercy Rule}
 A game will conclude once the game score shows a goal difference of 10.
 Ending the game is mandatory once a goal difference of 10 is reached.
 
+\subsection{Drop-Ball Rule}
+\label{sec:drop_ball_rule}
+
+In the event that there has been a referee error or a pause in game, and if a goal has not been scored by either team, the drop-ball rule shall apply. When resuming the game, the process should be the same as a normal start, however, none of the teams will be given kickoff, and no team would be allowed in the center circle until the \texttt{playing} state as there is no longer time allotted for kickoff. The only main example currently where the drop-ball rule would be required is when a referee timeout has been called with no goal being scored.
+
```
rvilling, line 524: This section needs to clarify a few things. 
- I presume a referee error needs to be acknowledged by the referee.
- What pauses in the game can occur? If it is just a referee timeout, then we should probably state that.
- What does "if a goal has not been scored by either team" mean? Does this mean at any point previously, or immediately beforehand, or within some particular time window?
- Does a drop ball mean that ball will always be placed in the center of the field as for a standard kickoff? This should be stated. It also seems like quite a severe reset in case of a referee error.
- If the ball is always placed in the center, where do robots start from? Is it from the sidelines? If so, does this mean that in case of any referee error or referee timeout is called, the robots stop playing and are placed on the sideline?

rules/game_process.tex
```diff
@@ -517,6 +517,16 @@ \subsection{Mercy Rule}
...
+\subsection{Ball Stop Rule}
+\label{sec:early_stopping}
+
+If a game half were to extend longer than the allotted time, due to an attack situation by a team, the half should end if the ball comes to a complete stop or when the ball leaves the field, with no immediate play by other robots.
```
ben47955, line 528: Maybe consider limiting the attack to penalty area instead of the field ?

rules/judgment.tex
```diff
@@ -80,7 +82,17 @@ \subsection{Referee--Team Communication}
...
+In this case, there should be a list of all the head and assistant referees that are willing to judge a game.
+The referees on the list should be voluntary but participation should be encouraged.
```
ben47955, line 96: Maybe specify this is offer to teams, but they can still choose their referees for there friendly games.

rules/judgment.tex
```diff
@@ -45,7 +45,9 @@ \subsection{Assistant Referees}
 An assistant referee will also put the robot back on the field.
 An assistant referee will also replace the ball when it goes off the field or becomes stuck between a players feet.
 
-The assistant referees can \textit{indicate} violations against the rules committed by robots to the head referee, so that the head referee can decide whether to penalize a certain robot or not.
+The head referee should talk to the assistant referees before the game starts to determine what tasks or lesser decisions they wish to relegate to them, however the final calls must come from the head referee.
+Examples include, assistant referees keeping track of indirect-kick rule compliance by a team in the event the head referee needs confirmation if a valid goal has been scored, indicate or to suggest a call for a specific violation (leaving the field, or motion in set) of the rules committed by the robots, to which the referee can act on accordingly.
```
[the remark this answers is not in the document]
TheJollyDuck: I have changed it to this better rewording.

TheJollyDuck:
I have added more clarifications as Rudi specified wrt to the Drop ball rule and the Aborting a Penalty. For the former, the ball and robot placement are now specified, as well as the consideration for when a ball has not been scored yet, which should be considered since the start of play at kick-off.

For the latter, the clarifications of not incrementing the penalty time and count is explicitly mentioned.

rules/judgment.tex
torchipeppo: Maybe also say that new referees seeking experience are particularly welcome? IIRC this was one of the intentions behind this.

rules/game_process.tex
TheJollyDuck: I thought of this as the last kickoff that was executed, regardless of whether it was the first half or after a timeout. In the case of a goal was previously scored, and that initiated a kickoff, I would say the conceding team would have kickoff.

arnemoos (approved):

torchipeppo (approved):

PR #166, opened 2024-12-02
Add standby to the initial kickoff procedure

torchipeppo:
Another tiny pull request for clarification.

The reason for this is twofold:

- While reviewing the drop-ball rule, I momentarily blanked on whether a restart after a timeout has Standby or not;
- In general, I noticed that there is no mention of the Standby state in either kickoff procedure: it's currently only in the specification of the robot states.

So, this little addition aims to better position the Standby state within the flow of the game, and hopefully provide an easy-to-find answer for any random robocupper wondering the same thing as I just did.

This is based on the GameController implementation, which does have a Standby state after a timeout.


TomJEllis:
Is this already mentioned in https://github.com/RoboCup-SPL/Rules/blob/master/rules/game_process.tex#L84-L87

TomJEllis (approved):

PR #170, opened 2024-12-13
First draft of the challenge with humanoid

ben47955:


Pass-Masters-Bridging-the-Leagues.tex
```diff
@@ -0,0 +1,206 @@
...
+One attacking robot is randomly selected to start with the ball.
+The defenders are positioned randomly D, E, or F.
+See Figure \ref{ch2:zone96}.
+In case of fallen robots, if the robot is not able to get up by itself, the team can request a fallen robot from the GameController.
+This will obviously result in a penalization for the robot and a possible time loss for the team.
+
+\section{Challenge Execution}
+For each attempt the attacking robots must be connected to GameController.
+Before each attempt, GameController is set to Initial.
+After the robots are positioned, the GameController changes to Ready state and then Set.
```
torchipeppo, line 171: Since the robots are manually placed, is there a necessity for the GameController to pass through the Ready state? We're probably going to need the GC to skip the SPL-specific Standby state, so if the answer is "yes" maybe we could arrange it to go directly from Initial to Set.

Pass-Masters-Bridging-the-Leagues.tex
```diff
@@ -0,0 +1,206 @@
...
+\centering
+\end{figure}
+
+\section{Field Setup}
+This challenge will be conducted on a standard SPL field.
+The challenge requires three virtual zones shown in Figure \ref{ch2:zone96}.
+The red zones represent area available to attacker robots. The blue zone represents the area where the defender robots are placed.
+The three zones should be made clearly visible to observers including the referee, by demarcating them on the field through the use of a green tape. The tape must not break the regular white field lines.
+Robots must localize using the regular filed lines.
+The ball used must be from the Humanoid league, featuring a dark-and-white patterned design.
```
torchipeppo, line 132:
```suggestion
The ball used must be from the Humanoid league, featuring a dark-and-white patterned design similar to that on a standard SPL ball.
```
Not the most pressing concern, but I would specify this unless there's a specific reason not to

Pass-Masters-Bridging-the-Leagues.tex
ben47955: Right, it would probably turn out in a Stanby of 0s. I think it should stay there for backward compatibility with regular games.

PR #178, opened 2025-04-07
Add suggestions for tasks to delegate + extra assistants

torchipeppo:


rules/judgment.tex
```diff
@@ -39,20 +39,36 @@ \subsection{Head Referee}
...
-The head referee should talk to the assistant referees before the game starts to determine what tasks or lesser decisions they wish to delegate to them, however the final calls must come from the head referee.
-Examples of delegated assistance to referee decisions include keeping track of indirect-kick rule compliance to help confirm if a valid goal has been scored and indicating possible rules violations requiring a penalty (such as leaving the field, or motion in set).
+At the discretion of the head referee, more tasks can be delegated to the assistants (\cref{sec:referee_delegation}).
 
 Assistant referees should only enter the field to execute a decision made by the main referee.
 They should not prevent robots from falling during the game.
 
+A game has at least two assistant referees. If agreed upon by the referee teams
+or under certain circumstances, additional assistants may be present,
+up to a total maximum of four. See \cref{sec:refSelection}.
```
torchipeppo, line 54: This maximum number is an initial guess, can be tweaked or removed entirely.

rules/judgment.tex
```diff
@@ -39,20 +39,36 @@ \subsection{Head Referee}
...
+
 The assistant referees should be equipped with a suitable referee jersey, and black or dark-blue socks.
 
+\subsubsection{The Standing Assistant}
+
+Up to one assistant referee may be designated as the \textit{standing assistant}.
```
torchipeppo, line 60: I'm open to naming suggestions as usual.
I was initially going to call this "gesture assistant," but since we expect them to be delegated the indirect kick rule more often than not, I tried to think of a name that doesn't directly correlate to any duties.

rules/app_competition_rules.tex
```diff
@@ -116,9 +116,21 @@ \subsection{Referee Selection and Requirements}
 Referees must have good knowledge of the rules as applied in the tournament, and the operator of the GameController must be experienced in using that software.
 Referees and the GameController should be selected among the more senior members of a team, and preferably have prior experience with games in the RoboCup Standard Platform league.
 
+If agreed upon by the referee teams or if scheduled by the Organizing Commettee,
```
arnemoos, line 119: committee

rules/judgment.tex
arnemoos: Since for some final day games we sometimes have 4 assistants I think this is a goog number.

rules/app_competition_rules.tex
torchipeppo: Fixed, ty

rules/judgment.tex
```diff
@@ -39,20 +39,36 @@ \subsection{Head Referee}
...
+This is especially encouraged if the game includes additional assistants
+beyond the standard two.
+
+If a \textit{standing assistant} is present, they are tasked with standing
+on the T-junction opposite to the technical area at all times and perform
+the gestures for the beginning of the \texttt{ready} phase and for set plays
+instead of the head referee.
+
+They may be delegated other tasks by the head referee, such as keeping track
```
arnemoos, line 69: I would be more in favor of strongly encouraging this as well because when the referee is doing the gesture it also knows that from now on the indirect kick counter ist reset.

rules/judgment.tex
```diff
@@ -66,6 +82,64 @@ \subsection{GameController Operator}
...
+        this could mean that assistants are allowed to pick up a robot that is
+        clearly intending to leave the carpeted area or going to get damaged
+        even if the head referee has not called ``Leaving the Field'' on it,
+        holding it until they get the attentiion of the head referee
+        and the call is made.
+        Alternatively, the assitant referees may be given full power to make
+        ``Leaving the Field'' calls in place of the head referee, but this solution
+        also requires the approval of the GameController operator since that means
+        they will have multiple people to pay attention to.
+  \item Indicating other violations requiring a penalty, with the call still
```
arnemoos, line 130: with the final call still

